PR #128, opened 2026-02-19
Update Gemini Embedding model

gratzc:
# Description

text-embedding-004 is deprecated/removed. gemini-embedding-001 is the current standard


gratzc:
Failed from no API keys for the run

PR #134, opened 2026-02-27
Add Mac vllm support for Docker Model Runner

### LINKED ISSUE #133: Add Mac vllm support for Docker Model Runner
## Summary

Docker Model Runner's new vLLM engine backend on Mac requires different HTTP handling than the existing llama.cpp backend. The `bx:http`/`cfhttp` HTTP client causes DMR's reverse proxy to drop the POST body when routing to vLLM, and vLLM does not yet support streaming responses through DMR.

## Detailed Description

Docker Model Runner recently added vLLM as a second inference engine alongside llama.cpp on macOS. Testing revealed two compatibility issues with the current `DockerModelRunnerService`:

### 1. POST body dropped by DMR reverse proxy

When `bx:http` sends a POST request, it uses HTTP behaviors (likely `Expect: 100-continue` or `Transfer-Encoding: chunked`) that cause DMR's reverse proxy to silently drop the request body before forwarding to the vLLM engine. The vLLM backend then returns a 400/422 error because it receives an empty payload. This does **not** affect llama.cpp — only vLLM.

Java's `HttpURLConnection` sends the body inline in a single write and works correctly with both backends.

### 2. vLLM ignores `stream=true` through DMR

When `stream=true` is set in the request payload, llama.cpp correctly returns an SSE stream. vLLM through DMR currently ignores this flag and returns a standard JSON response (non-streaming `chat.completion` object instead of `chat.completion.chunk` SSE events).

This means `chatStream()` receives a plain JSON body instead of SSE data, causing the stream parser to produce no output — the callback is never invoked and the user gets an empty response.

### Replace `bx:http` with Java `HttpURLConnection` for DMR HTTP calls

Override `sendChatRequest()` and `sendStreamRequest()` in `DockerModelRunnerService` to use `java.net.HttpURLConnection` directly, bypassing `bx:http` entirely. This keeps the fix scoped to DMR without affecting other providers.

Helper methods needed:
- `javaHttpPost()` — executes a POST with proper auth/headers/timeout, returns `{ statusCode, body, headers }`
- `readStream()` — reads a `java.io.InputStream` fully into a string

### Handle vLLM non-streaming fallback in `sendStreamRequest()`

When reading the SSE stream, if no `data:` lines are found but the response body is valid JSON, treat it as a non-streaming response from vLLM:

1. Accumulate non-SSE lines into a buffer while reading
2. After the stream ends, if no SSE events were parsed, check the buffer for a JSON object
3. Convert the `choices[].message` format to a synthetic `choices[].delta` streaming chunk
4. Pass both `content` and `tool_calls` through to the callback (vLLM may return tool call responses)
5. Log the fallback so it's diagnosable in production

dandyfopper:
# Description

Docker Model Runner recently added vLLM as a second inference engine alongside llama.cpp on macOS. Testing revealed that `bx:http`/`cfhttp` causes DMR's reverse proxy to silently drop the POST body when routing to vLLM (likely due to `Expect: 100-continue` or `Transfer-Encoding: chunked` handling). Additionally, vLLM through DMR does not yet support streaming — it returns a standard JSON response even when `stream=true` is requested.

This PR replaces `bx:http` with Java `HttpURLConnection` for all Docker Model Runner HTTP calls and adds a fallback that converts non-streaming vLLM responses into synthetic streaming chunks so `chatStream()` works transparently with both backends.

### Changes

- **Override `sendChatRequest()`** with Java `HttpURLConnection` implementation that bypasses `bx:http`
- **Override `sendStreamRequest()`** with Java-based SSE parser that reads the stream line-by-line
- **vLLM non-streaming fallback**: when no SSE events are found in the response, detects a plain JSON body and converts `choices[].message` to a synthetic `choices[].delta` streaming chunk, including both `content` and `tool_calls`
- **`javaHttpPost()` helper**: executes POST with auth/headers/timeout, returns `{ statusCode, body, headers }`
- **`readStream()` helper**: reads a `java.io.InputStream` fully into a string
- **Null-safe `getErrorStream()`**: guards against `HttpURLConnection.getErrorStream()` returning null on empty error bodies (e.g., bare 503)
- **Null-safe connection cleanup**: initializes `conn` with `javacast("null", "")` and checks `isNull(conn)` before `disconnect()` in catch blocks
- **SSE buffer newlines**: inserts `chr(10)` between accumulated non-SSE lines so pretty-printed JSON reassembles correctly

All changes are scoped to `DockerModelRunnerService` — no other providers are affected.

## Jira/Github Issues

Closes https://github.com/ortus-boxlang/bx-ai/issues/133

## Type of change

-   [ ] Bug Fix
-   [ ] Improvement
-   [x] New Feature
-   [x] This change requires a documentation update

## Checklist

-   [ ] I have added tests that prove my fix is effective or that my feature works


lmajano:
I prefer we stick with http support in BoxLang. 

If we have a bug or a feature we need to add to our http caller then let’s do that. I do not want to mix implementations. 

What exactly do you need the http component to support. I need a reproducible case too.

dandyfopper:
@lmajano a good point, I dug deeper and it turns out it is an issue with downgrading bx:http default http/2 to http/1.1

### Root Cause

`bx:http` defaults to `httpVersion="HTTP/2"`. Java's `HttpClient` then attempts an HTTP/2 cleartext upgrade (h2c) over the plain HTTP connection to Docker Model Runner. DMR's reverse proxy drops the POST body during protocol translation when routing to the vLLM backend. The llama.cpp backend is unaffected.

**Proof** — tested locally with a minimal Java program hitting DMR directly:

```HttpClient.Version.HTTP_2 → vLLM: 422 {"detail":[{"type":"missing","loc":["body"],"msg":"Field required"}]}
HttpClient.Version.HTTP_1_1 → vLLM: 200 (works)
HttpClient.Version.HTTP_2 → llama.cpp: 200 (works, unaffected)
```

Then confirmed with actual `bx:http`:
bx:http httpVersion="HTTP/1.1" → vLLM: 200 ✅
bx:http (default HTTP/2) → vLLM: 422 ❌

This is a Docker Model Runner proxy bug — I'll file an issue with Docker. But the workaround is trivial.
https://github.com/docker/model-runner/issues/716

### Fix

Override `sendChatRequest()` and `sendStreamRequest()` in `DockerModelRunnerService` with copies that add `httpVersion="HTTP/1.1"` (tag syntax) and `.http1()` (fluent API). Also added it to the native DMR API methods (`listModels`, `pullModel`, `getModelInfo`, `isAvailable`).

For streaming, vLLM also ignores `stream=true` through DMR and returns a plain `chat.completion` JSON response instead of SSE. Added a `sendNonStreamingFallback()` method that detects zero SSE chunks, re-sends as non-streaming, and synthesizes a streaming chunk for the callback.

This is probably due to the new nature of the feature and streaming will be included in docker for vllm shortly - the pull request anticipates this and will still work when streaming is enabled.

No new dependencies, no mixed implementations — all `bx:http`.

I'll update the PR.

dandyfopper:
looks like this is really a docker model issue not bx-ai - docker seem to be reacting to fixing it so not appropriate to change bx-ai

lmajano:
great!

Issue #105, opened 2026-01-09
Create the plan for image models


lmajano:

crismantilla:
Here's the proposed implementation design for allowing I/O of images to LLM models. Please let me know of any feedback you have, and if we're good to continue with this implementation.

[AiImage_Implementation_Document.docx](https://github.com/user-attachments/files/25354580/AiImage_Implementation_Document.docx)

crismantilla:
Here's the implementation in markdown.
*Ortus Solutions — Implementation Document*

**IMPLEMENTATION TEMPLATE DOCUMENT** *\[bx-ai / Image I/O\]*

| **Project Name** | *bx-ai* |
|----|----|
| **Repository URL** | [*<u>https://github.com/ortus-boxlang/bx-ai</u>*](https://github.com/ortus-boxlang/bx-ai) |
| **Issue / Ticket \#** | [*<u>https://github.com/ortus-boxlang/bx-ai/issues/105</u>*](https://github.com/ortus-boxlang/bx-ai/issues/105) |
| **Team Members** | *Cristian Mantilla (System Architect), Christian Marin (AI Integration Lead), Wesley Brutus (API Researcher & Reviewer), Areesh Chowdhury (System Implementation)* |
| **Date Created** | *02/12/2026* |
| **Target Completion** | *02/28/2026* |
| **Status** | *Draft* |
| **Reviewed By** | *\[Engineer/Mentor name\]* |

**1. Objectives & Goals**

**1.1 Problem Statement**

**Problem: Current AI library functions do not support image input, and
image output is inconvenient to achieve.**

**1.2 Objectives**

1\. Implement Image I/O into Boxlang’s LLM Library functions for
providers that support it.

2\. Implement provider-specific API parameters and options into the AI
Library as a function.

**1.3 Success Criteria**

• Working integration of LLM-based image I/O functionalities into
Boxlang’s AI library.

• Code review approved.

• Documentation updated to reflect changes

**2. Scope**

**2.1 In Scope**

Implementation of a **new image-focused returnFormat** that returns
**only a base64 string** from image generation responses, specifically
parsing OpenAI responses where the image is found under output →
image_generation_call → result\
Planning Document

Extension of **aiChat()** to:

- Accept image inputs via a new images key

- Support receiving image outputs when returnFormat: "image" is
  specified

Automatic assumption of params: { tools: \[{ type: "image_generation"
}\] } when using OpenAI with returnFormat: "image" and no explicit tool
[...]

PR #147, opened 2026-03-11
Add Image I/O System, includes Image I/O with OpenAI Provider

crismantilla:
# Description
This pull requests addresses Issue https://github.com/ortus-boxlang/bx-ai/issues/105 regarding implementing Image I/O within AI Providers.

Three new features have been created for the module. These are:
- Parsing of an images array when inputted into the options array parameter
- A new returnFormat "images" that returns the base64 of the generated image.
- A new wrapper function aiImage() that by default returns the base64 of the generated image, and has a optional parameter to store the image to disk at a specified path.

The features are parsed by the provider classes through the supportsImageGeneration() and supportsImageInput() functions, which individual providers can override whenever image generation logic is implemented for them (they default to false if not ovverriden from the BaseService class). 

To accompany these features, Image I/O support was implemented for OpenAI. In order to achieve Image I/O, the OpenAIService class contains logic for switching from the Chat API to the Responses API when Image I/O is detected, which was necessary as the Chat API did not support Image I/O. 
When returnFormat "image" is utilized, a "image_generation" tool is injected into the datapacket before sending the request to trigger OpenAI's Built-in tool.

Finally, tests in the OpenAI unit test were added to test for Image I/O.

All test cases within the OpenAI unit test pass.
Running all tests returns 87 failed, however, running all tests from before any of the PR changes have 88 fail (Note: For both runs, only the OpenAI key was available). These failures are not a result of the PR changes. 

## Jira/Github Issues
BoxLang Jira: N/A
Module Issues: https://github.com/ortus-boxlang/bx-ai/issues/105

## Type of change
-   [x] New Feature
-   [x] This change requires a documentation update
In specific, the README may want to display the ability for Image I/O.

## Checklist

-   [?] My code follows the style guidelines of this project [cfformat](../.cfformat.json) and [Java](../ortus-java-style.xml)
NOTE: running box run-script format returns a message saying "The script [format] does not exist in this package." 
Therefore, I cannot verify if the .ccformat.json is followed.
However, ./gradlew spotlessCheck (for Java style guidelines) ran and passed.

-   [] I have made corresponding changes to the documentation
NOTE: It is unclear whether the README should reflect these new features (Especially since only OpenAI has Image I/O Implementation) . Do reply if so, then we will get right on that.

### DIFF
--- a/src/main/bx/bifs/aiChat.bx
+++ b/src/main/bx/bifs/aiChat.bx
@@ -18,7 +18,7 @@
 class {
 
 	static {
-		final VALID_RESPONSE_FORMATS = [ "single", "all", "raw", "json", "xml" ]
+		final VALID_RESPONSE_FORMATS = [ "single", "all", "raw", "json", "xml", "image" ]
 		final DEFAULT_OPTIONS = {
 			returnFormat : "single",  // aiChat() defaults to single for convenience
 		}
@@ -50,7 +50,8 @@ class {
 	 * <li>logResponseToConsole:boolean - Log the response to the console. The default is false</li>
 	 * <li>logRequest:boolean - Log the request into the ai.log. The default is false</li>
 	 * <li>logRequestToConsole:boolean - Log the request to the console. The default is false</li>
-	 * <li>returnFormat:string - The return format of the response. The default is "single" message. Valid formats are: "single", "all", "raw", "json", "xml"</li>
+	 * <li>returnFormat:string - The return format of the response. The default is "single" message. Valid formats are: "single", "all", "raw", "json", "xml", "image"</li>
+	 * <li>images:array - Optional base64-encoded images to include in the request (OpenAI Response API mode).</li>
 	 * </ul>
 	 *
 	 * @messages The messages to pass into the required model. This depends on the provider. It can be a simple string, a struct representing a message or an array of messages or an actual AiMessage object
--- a/src/main/bx/bifs/aiChat.bx
+++ b/src/main/bx/bifs/aiChat.bx
@@ -51,7 +51,7 @@ class {
 	 * <li>logRequest:boolean - Log the request into the ai.log. The default is false</li>
 	 * <li>logRequestToConsole:boolean - Log the request to the console. The default is false</li>
 	 * <li>returnFormat:string - The return format of the response. The default is "single" message. Valid formats are: "single", "all", "raw", "json", "xml", "image"</li>
-	 * <li>images:array - Optional base64-encoded images to include in the request (OpenAI Response API mode).</li>
+	 * <li>images:array - Optional base64-encoded images to include in the request.</li>
 	 * </ul>
 	 *
 	 * @messages The messages to pass into the required model. This depends on the provider. It can be a simple string, a struct representing a message or an array of messages or an actual AiMessage object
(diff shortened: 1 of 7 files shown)


src/main/bx/bifs/aiChat.bx
```diff
@@ -50,7 +50,8 @@ class {
 	 * <li>logResponseToConsole:boolean - Log the response to the console. The default is false</li>
 	 * <li>logRequest:boolean - Log the request into the ai.log. The default is false</li>
 	 * <li>logRequestToConsole:boolean - Log the request to the console. The default is false</li>
-	 * <li>returnFormat:string - The return format of the response. The default is "single" message. Valid formats are: "single", "all", "raw", "json", "xml"</li>
+	 * <li>returnFormat:string - The return format of the response. The default is "single" message. Valid formats are: "single", "all", "raw", "json", "xml", "image"</li>
+	 * <li>images:array - Optional base64-encoded images to include in the request.</li>
```
lmajano, line 54: This is already addressed by using the `AiMessage()` image inputs. I prefer we use the messages object for more complex interactions, than add yet another option.

src/main/bx/bifs/aiChatAsync.bx
```diff
@@ -42,7 +42,8 @@ class extends="aiChat"{
 	 * <li>logResponseToConsole:boolean - Log the response to the console. The default is false</li>
 	 * <li>logRequest:boolean - Log the request into the ai.log. The default is false</li>
 	 * <li>logRequestToConsole:boolean - Log the request to the console. The default is false</li>
-	 * <li>returnFormat:string - The return format of the response. The default is "single" message. Valid formats are: "single", "all", "raw", "json", "xml"</li>
+	* <li>returnFormat:string - The return format of the response. The default is "single" message. Valid formats are: "single", "all", "raw", "json", "xml", "image"</li>
+	 * <li>images:array - Optional base64-encoded images to include in the request.</li>
```
lmajano, line 46: This is already addressed by using the AiMessage() image inputs. I prefer we use the messages object for more complex interactions, than add yet another option.

src/main/bx/models/requests/AiChatRequest.bx
```diff
@@ -43,6 +43,11 @@ class extends="AiBaseRequest"{
 	 */
 	property name="structuredOutputDefinition" type="any" default={};
 
+	/**
```
lmajano, line 46: This is already addressed by using the AiMessage() image inputs. I prefer we use the messages object for more complex interactions, than add yet another option.

lmajano:
This repo has now conflicts because our entire architecture changed to support multi-model requests.

Please see that now we have a capabilities folder in the providers, so we can add image capabities to specific providers.  Please review the new architecture, and adapt your code.

We have to think that `aiImage()` would produce an image.  The question is if we want `aiChat()` to do the same?

crismantilla:
> This repo has now conflicts because our entire architecture changed to support multi-model requests.
> 
> Please see that now we have a capabilities folder in the providers, so we can add image capabities to specific providers. Please review the new architecture, and adapt your code.
> 
> We have to think that `aiImage()` would produce an image. The question is if we want `aiChat()` to do the same?

No problem, I'll modify the repo to align with the new architecture and get back to you. I'll make sure it utilizes both AiMessage() and the capabilities system.

As for your question, I do agree aiChat() may not need Image I/O capabilities with the existence of aiImage(). I saw that there was no way to invoke the built-in OpenAI tool and added the image returnFormat.
However, aiImage() would fulfill that same purpose without the need of the extra option. Additionally, implementing aiImage() not as a aiChat() wrapper function allows for more desired behavior, such as an option to return both the image and the text response from the model.

Thank you for your reviews and suggestions, changes will be made soon.

PR #175, opened 2026-04-19
feat(mcp): fill observability and analytics gaps in MCP server

lmajano:
- Fix thread-safety: byMethod, byTool, byUri, byName, byCode struct
  mutations are now wrapped in named locks (previously unsynchronized
  concurrent read-modify-writes could silently lose updates)

- Track security failures: basic auth, API key, and body-size violations
  now increment dedicated AtomicInteger counters in MCPServerStats and are
  visible in getStats()/getSummary() under the new "security" key;
  MCPServer exposes recordSecurityFailure() for processor delegation

- Record paused-request stats: handleRequest() now calls recordRequest()
  with SERVER_PAUSED before the early return instead of silently dropping
  the request from stats

- Announce onMCPError for METHOD_NOT_FOUND: the default switch case was
  the only error path that never fired the BoxAnnounce event; fixed

- Track per-tool errors: handleToolCall() wraps invoke() in try/catch and
  calls recordToolError() before rethrowing; MCPServerStats gains
  byTool[name].errors and errors.byTool roll-up counters

- Track active concurrent requests: MCPServerStats gains an activeRequests
  AtomicInteger; handleRequest() increments on entry and decrements in a
  finally block; exposed in getStats() and getSummary()

- Add requestsPerMinute to getSummary() for throughput visibility

- Propagate X-Request-ID: HTTPTransport reads the incoming header (or
  generates a UUID if absent); StdioTransport always generates one;
  processWithTransport() echoes it as X-Request-ID in the response and
  includes it in onMCPRequest/onMCPResponse event payloads

https://claude.ai/code/session_01LwwePeKYcECkZk8PPtXCWn

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### 🥊 New Features
 
+- **MCP Server Observability & Analytics Improvements**: Multiple gaps in the MCP server's observability and analytics have been addressed.
+  - **Thread-safety fix**: `byMethod`, `byTool`, `byUri`, `byName`, and `byCode` counters in `MCPServerStats` were plain struct mutations happening outside any lock, causing silent lost updates under concurrent load. All are now wrapped in dedicated named locks.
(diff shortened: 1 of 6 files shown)


lmajano:
@copilot resolve the merge conflicts in this pull request

Issue #176, opened 2026-04-20
ClosureTool rejects struct/array values for string-typed params, breaking MCP tools with JSON object arguments


dougcain:
## What are the steps to reproduce this issue?

1. Register an AI tool whose callable declares a parameter with an explicit `string` type (e.g. a tool that accepts a JSON-encoded struct such as `source_context`, `key_decisions`, or any `Record<string, unknown>` arg from an MCP schema):
   ```bx
   aiTool(
       "captureHandoff",
       "Capture a handoff",
       ( required string summary, string source_context = "" ) => { ... }
   )
   ```
2. Call the tool from an MCP client (Claude Code, Claude Desktop, Cursor, etc.) sending the JSON-typed param as an **actual JSON object** in the `arguments` payload, not as a serialized string:
   ```json
   {
     "method": "tools/call",
     "params": {
       "name": "captureHandoff",
       "arguments": {
         "summary": "...",
         "source_context": { "repo": "github.com/acme/app", "agent_id": "claude-code" }
       }
     }
   }
   ```
3. Observe the Lambda / server logs.

## What happens?

The call fails before the callable body runs with:

```
Can't cast Struct<Ordered> to a string
```

`ClosureTool.doInvoke()` passes `arguments.args` straight to the callable via `argumentCollection`, and BoxLang's typed-arg binding rejects the `Struct<Ordered>` value where the declared parameter type is `string`. The user-facing result is that every affected tool call 500s, with no way to handle the payload from inside the callable.

## What were you expecting to happen?

The tool should receive the JSON-encoded object as a string, letting the callable deserialize it (`jsonDeserialize( source_context )`). MCP tool schemas generated by `ClosureTool.getArgumentsSchema()` declare every parameter as `"type": "string"`, so clients sending raw JSON objects are a legitimate (and common) wire-format variation — the schema and the runtime should agree.

## Any logs, error output, etc?

```
[ERROR] [AI] om_capture_handoff failed: Can't cast Struct<Ordered> to a string.
```

Reproducible on both `bxai@3.0.0+10` and the current `bxai@3.1.0+11`.

## Any other comments?

Proposed fix: in `ClosureTool.doInvoke()`, before invoking the callable, walk `variables.callable.$bx.meta.parameters` and, for any parameter declared as `string` whose incoming value is a non-simple type, `jsonSerialize()` it so the existing typed-arg binding succeeds.

[...]

PR #177, opened 2026-04-20
fix(tools): coerce struct/array args to JSON for string-typed params …

dandyfopper:
# Description

`ClosureTool.doInvoke()` currently passes `arguments.args` straight into the callable via `argumentCollection`. When an MCP client sends a JSON-typed field (e.g. `source_context`, `key_decisions`, `files_modified`) as an actual JSON object or array — which several popular clients and SDKs do — BoxLang's typed-arg binding rejects the `Struct<Ordered>` / `Array` value against the declared `string` parameter and throws `Can't cast Struct<Ordered> to a string` before the callable body runs. Every affected tool call 500s with no opportunity for the callable to handle the payload.

This patch walks the callable's declared parameters and, for any parameter whose declared type is `string` and whose incoming value is non-simple, `jsonSerialize()` the value before delegating. The callable then deserializes as it already would (`jsonDeserialize( source_context )`) with no change, and string args remain untouched.

The generated MCP schema declares every parameter as `"type": "string"`, so the runtime now consistently accepts both supported wire formats (stringified JSON and raw JSON objects/arrays). Callables that explicitly declare `struct` / `array` / `any` params are unaffected — coercion only runs when the declared type is `string`.

## Jira/Github Issues

Module Issues: https://github.com/ortus-boxlang/bx-ai/issues/176

## Type of change

- [x] Bug Fix

## Checklist

- [x] My code follows the style guidelines of this project [cfformat](../.cfformat.json) and [Java](../ortus-java-style.xml)
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/main/bx/models/tools/ClosureTool.bx
+++ b/src/main/bx/models/tools/ClosureTool.bx
@@ -78,6 +78,19 @@ class extends="BaseTool" {
 				message : "No callable has been set on this ClosureTool. Use call() before invoking."
 			)
 		}
+		// Coerce structured values to JSON strings for string-typed params. MCP clients may send
+		// object/array payloads for params whose schema declares "string" (e.g. JSON-encoded fields
+		// like source_context or key_decisions); without this, BoxLang's typed-arg binding would
+		// throw "Can't cast Struct<Ordered> to a string" before the callable runs.
+		var incomingArgs = arguments.args
+		variables.callable.$bx.meta.parameters.each( param => {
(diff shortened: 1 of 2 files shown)


lmajano:
@copilot create a new pr on this to generate the changelog for the fix

PR #223, opened 2026-08-01
feat: HITL extraction (Phase 2) — approval policies, coordinator, gateway-attached HITL

lmajano:
# Description

Adds Phase 2 of the BX AI Gateway & HITL Generalization plan — extracts the approval-decision and presentation logic out of `HumanInTheLoopMiddleware` into reusable, testable collaborators, and wires the middleware up to the Phase 1 Gateway SPI (#220) so any gateway (not just the built-in CLI/web behavior) can present an approval request and resolve it.

This is purely additive plus one in-place refactor with **zero behavior change** for existing usage (`toolsRequiringApproval`/`mode: "cli"`/`mode: "web"` with no `policy`/`gateway` supplied) — every pre-existing HITL unit test passes unmodified.

**What's included:**
- New `models/hitl/` package:
  - `IApprovalPolicy` — decides whether a pending tool call needs human approval
  - `ToolNameApprovalPolicy` — the original tool-name-list behavior, now a standalone policy
  - `AnnotationApprovalPolicy` — approval driven by a `@requiresApproval` annotation on a tool's `doInvoke()`, mirroring the annotation-scanning pattern `MCPServer.scanClass()` already uses
  - `RiskLevelApprovalPolicy` — approval driven by a `@riskLevel(...)` annotation vs. a configured threshold
  - `CallbackApprovalPolicy` — a custom `function( context ) => boolean` decision
  - `CompositeApprovalPolicy` — combines several policies with `any`/`all` semantics
  - `HumanInteractionCoordinator` — owns the suspend-for-approval lifecycle against an attached gateway: creates and tracks an `AgentSuspension`, and resolves it exactly once via an atomic compare-and-set (`java.util.concurrent.ConcurrentHashMap.replace()`) so two simultaneous decisions for the same suspension can't both win. Also validates a human's edited arguments against the tool's declared schema before accepting an "edit" decision — an invalid edit is normalized into a rejection with a reason (not thrown, not re-suspended).
- `HumanInTheLoopMiddleware.bx` refactored into a thin adapter over the above: new optional `policy` and `gateway` constructor arguments. With a `gateway` attached, presentation is delegated through the coordinator instead of the built-in CLI prompt / bare suspend. With neither supplied, behavior is byte-for-byte the same as before. The previously dead `approvalCallback` constructor argument is now wired up as a `CallbackApprovalPolicy`.

[...]

PR #224, opened 2026-08-01
feat: CLI gateway + compatibility cleanup (Phase 3)

lmajano:
# Description

Adds Phase 3 of the BX AI Gateway & HITL Generalization plan — extracts the blocking stdin/stdout CLI approval prompt out of `HumanInTheLoopMiddleware` into a real `IGateway` implementation, so the CLI is just another gateway rather than a special case hardcoded into the middleware.

This is **zero behavior change** for all existing `toolsRequiringApproval`/`mode` usage — every pre-existing HITL/suspend-resume test passes unmodified — plus two small, deliberate cleanups riding along.

**What's included:**
- `models/gateway/cli/CliGateway.bx` — the canonical/reference `IGateway`: wraps the original prompt banner and A/R/Q parsing verbatim, resolving synchronously (`requestHumanInteraction()` always returns a decision immediately). Its input-reading mechanism is injectable (`inputReader` constructor arg, defaulting to `cliRead()`) purely for testability — no real stdin needed in tests.
- `HumanInTheLoopMiddleware.bx`: the private `cliApprove()` method is gone. The constructor now always resolves a `gateway` — either the one explicitly passed, or (mode `"cli"`, still the default) a freshly-constructed `CliGateway`. Mode `"web"` with no gateway supplied is the only case left with no gateway attached, and suspends exactly as before.
- **Cleanup #1**: unrecognized CLI input used to silently cancel the run. `CliGateway` now re-prompts up to 3 times before falling back to cancel.
- **Cleanup #2**: an unrecognized `mode` string (a typo, e.g.) used to silently fall through to CLI behavior with no signal. It still falls back to a `CliGateway`, but now prints a `[DEPRECATED]` console warning pointing at the new `gateway:` constructor argument.

**Testing:** `CliGatewayTest.java` (approve/reject/quit/re-prompt-then-succeed/re-prompt-exhaustion, all via the injected `inputReader`, no real stdin touched) plus new compatibility tests proving: a default-constructed `HumanInTheLoopMiddleware` auto-attaches a `CliGateway`; explicit `mode: "cli"` does too; `mode: "web"` still has no gateway attached and still suspends; an unrecognized mode still falls back safely. All pre-existing HITL and `SuspendResumeIntegrationTest` tests pass unmodified. Verified zero regressions via a before/after full-suite diff (161 pre-existing/environment-dependent failures identical on both sides — live API keys, network, DB — none related to this change).

Builds on Phase 1 (#220) and Phase 2 (#223, currently in review).

## Jira/Github Issues

N/A — no tracking issue was provided for this work.

## Type of change

-   [x] New Feature
-   [ ] Bug Fix
-   [ ] Improvement

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01PtqqU4LCQ2oVwgeFDLJj33)_


lmajano:
CI shows failures across a couple of the parallel test runs on this commit, but all of them are live-provider flakes unrelated to this PR's diff (which only touches `models/gateway/cli/CliGateway.bx` and `HumanInTheLoopMiddleware.bx` — no provider code):

- `OllamaTest > Test Ollama Tools` — failed with a network timeout in one run, and with an unexpected/wrong live-model tool-call answer in another (`expected to contain a match for: salvador|90` but got a different city).
- `CohereTest > Test Cohere streaming chat` — failed on an unexpected live streaming response (`expected to contain: 42` but got `as a`).

Both are live external-API calls, not deterministic assertions on code this PR changed. Every HITL/gateway/CliGateway test passes across all runs. I don't have permission to re-run the failed jobs via the API (403) — flagging here for a maintainer re-run, and I'll keep watching for the recovery.

---

PR #225, opened 2026-08-02
feat: generic HTTP/webhook gateway (Phase 4)

lmajano:
# Description

Adds Phase 4 of the BX AI Gateway & HITL Generalization plan — a generic, network-reachable `IGateway` implementation exposing inbound events and human-in-the-loop interactions over signed HTTP. This is the first gateway in the series with real network exposure (CLI/Mock are purely in-process), so security is front-loaded into this phase rather than deferred.

**What's included:**
- `models/gateway/http/GatewaySecurity.bx` — HMAC-SHA256 request signing/verification. `sign()`/`verify()` compute/check a signature over `timestamp.nonce.rawBody`, with timestamp-tolerance freshness checking and constant-time comparison (delegating to `java.security.MessageDigest.isEqual()`, the JDK's own timing-safe byte comparison).
- `models/gateway/http/HttpGateway.bx` — implements `IGateway`'s inbound/outbound/human-in-the-loop capabilities:
  - `verifyInbound()`/`parseInbound()` — signed inbound platform events, normalized into `GatewayMessage`
  - `deliver()` — POSTs to a configured `callbackUrl` if set, always records locally for polling/observability
  - `requestHumanInteraction()`/`parseHumanDecision()` — always resolves asynchronously (HTTP has no synchronous channel to a human); pending interactions are tracked with a configurable TTL (`requestTTLSeconds`, default 900s)
  - `resolveInteraction()` — atomic claim via `java.util.concurrent.ConcurrentHashMap.replace()`, so a duplicate decision POST for an already-resolved interaction throws rather than silently overwriting it
  - Nonce replay protection: every verified request's nonce is tracked and rejected if reused
- `models/gateway/http/GatewayRequestProcessor.bx` — the transport-agnostic front controller (reuses `mcp/transports/HTTPTransport` directly rather than subclassing it — CGI/CORS/security-header handling is already generic enough), mirroring `MCPRequestProcessor`'s `processHttp()` shape but routing REST-style:
  - `POST /bxai/gateways/{gatewayName}/events` — inbound platform event
  - `GET /bxai/interactions/{requestID}` — poll a pending interaction
  - `POST /bxai/interactions/{requestID}/decisions` — submit a human's decision
  - Gateways resolve via `gatewayRegistry()` (an application registers its configured `HttpGateway` instance(s) at startup) rather than via the URL, keeping secrets out of the request path.
- `public/gateway.bxm` — the HTTP entry point, mirroring `public/mcp.bxm`'s front-controller pattern.
- `aiGateway()` now also resolves `"cli"` (previously only constructible directly as `new CliGateway()`, an oversight from Phase 3) alongside the new `"http"`.

**Design note (documented in `HttpGateway.bx`'s header):** this gateway does not itself call `AiAgent.resume()` — that requires knowing which agent/checkpointer an interaction belongs to, which is application state this module doesn't own. `GatewayContext.threadID`/`agentID`, set by the caller when presenting a request, are persisted alongside the pending interaction and returned by both the GET and decision-POST responses, so the application can correlate the resolved decision back to `agent.resume(decision, threadID, editedData)` itself.

**Testing:** `HttpGatewayTest` (capabilities, signing round-trip, tampered-body rejection, interaction lifecycle + atomic resolve) plus `GatewayRequestProcessorTest` — real HTTP-level round-trip tests using the `bx-web-support` mock HTTP exchange (the same technique `mcpServerTest.java` uses for `MCPRequestProcessor`), covering the full exit criterion: suspend → GET pending → signed POST decision → resolved decision returned; duplicate decision rejected (409); expired interaction rejected (410); missing/bad signature rejected (401); nonce replay rejected (401); signed inbound event verified and parsed. Verified zero regressions via a before/after full-suite diff (161 pre-existing/environment-dependent failures identical on both sides — live API keys, network, DB — none related to this change).

Builds on Phases 1–3 (#220, #223, #224), all merged into `development`.

## Jira/Github Issues

N/A — no tracking issue was provided for this work.

## Type of change

-   [x] New Feature
-   [ ] Bug Fix
-   [ ] Improvement

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01PtqqU4LCQ2oVwgeFDLJj33)_


lmajano:
The CI failure on check `91442911508` (`tests / Tests bx-ai`) is unrelated to this PR's diff — investigated the job logs:

- `HuggingFaceTest > Test HuggingFace AI chat` — failed on a live API quota error: *"You have depleted your monthly included credits. Purchase pre-paid credits to continue using Inference Providers."*
- `OllamaTest > Should handle custom model parameter for Ollama` — failed on a live-model response-content assertion mismatch (`expected to contain a match for: [4four] but was: 2`).

Both are live-provider/network-dependent tests; nothing in this PR touches `HuggingFaceService.bx` or `OllamaService.bx`. Confirming these are flakes and not a regression: two other parallel CI runs against the identical commit (`91442881516`, `91442872478`) both completed with 0 failures.

No fix needed — will keep an eye out in case it recurs on the next push.

---

PR #229, opened 2026-08-02
fix: route returnFormat "json" through JSONExtractorTransformer so prose-wrapped JSON no longer throws

### LINKED ISSUE #222: asJson() throws when a reply contains valid JSON plus any surrounding text — JSONExtractorTransformer exists but is never wired in
## What are the steps to reproduce this issue?

1. Ask any provider for JSON via `.asJson()` (i.e. `returnFormat: "json"`):

   ```js
   var verdict = aiMessage()
       .system( "Reply STRICTLY as JSON: {""related"": true|false}. No prose." )
       .user( "Are these two notes related? …" )
       .to( aiModel( provider: "bedrock", params: {
            "model": "eu.anthropic.claude-sonnet-4-6", "max_tokens": 256
       } ) )
       .asJson()
       .run();
   ```

2. Have the model reply with a valid JSON object plus **anything else**. Any of
   these is enough:

   ```
   Here you go: {"related": true}
   ```
   ```
   {"related": true}

   Wait, let me reconsider.

   {"related": false}
   ```
   ```
   {"related": true}
   Let me know if you'd like the reasoning.
   ```

3. The call throws.

## What happens?

`jsonDeserialize` throws on the full reply string and the call is lost — no
value, no partial result, no fallback. The valid object that *is* in the reply
is never reached.

`case "json"` passes the entire content to `jsonDeserialize()` after only
stripping a markdown fence:

```js
case "json":
    return jsonDeserialize( extractFromCodeBlock( content, "json" ) );
```

`extractFromCodeBlock` handles ```` ```json ```` fences and nothing else — no
brace matching, no scan for the JSON within the text. So the deserialize only
succeeds when the model's reply is *exactly* one JSON value end to end.

This is **not provider-specific** — the same line appears in every provider that
implements the switch, so every one of them fails the same way:

| Provider | Line |
|---|---|
| `BedrockService.bx` | 873 |
| `ClaudeService.bx` | 561 |
| `OpenAIService.bx` | 241 |
| `GeminiService.bx` | 257 |
| `CohereService.bx` | 368 |

…plus everything inheriting the OpenAI path (`OpenAICompatibleService` and its
subclasses).

**What makes this look like an oversight rather than a design choice**: the
module already ships `models/transformers/JSONExtractorTransformer.bx`, which
does precisely the tolerant extraction needed here — strips markdown fences,
finds the JSON inside mixed text via `extractBracedContent()` (a string-aware,
escape-aware brace/bracket scan), validates with `isJSON()` before parsing, and
has a `strictMode` flag defaulting to `false` so a bad parse degrades to `{}`
instead of throwing. It is a complete solution to this exact problem — and
nothing wires it to `returnFormat: "json"`. Grepping the module,
`JSONExtractorTransformer` has no caller on the `asJson()` path at all.

## What were you expecting to happen?

`.asJson()` to return the JSON object the model produced, as long as the reply
contains one — the way `JSONExtractorTransformer` already does.

Concretely, `case "json"` routing through the transformer instead of calling
`jsonDeserialize` directly:

```js
case "json":
    return new models.transformers.JSONExtractorTransformer( { strictMode : false } )
        .transform( content );
```

That keeps the currently-working case working (a clean reply is returned by
`findJSONInText`'s `isJSON( content )` fast path, unchanged), and additionally
handles prose-wrapped JSON, which today throws.

Two sub-decisions worth your call rather than ours:

- **`strictMode`.** `false` returns `{}` on an unusable reply; `true` throws a
  typed `JSONExtractor.InvalidJSON` / `JSONExtractor.NoJSONFound`. Either beats
  the current raw `jsonDeserialize` error, since both tell the caller which
  layer failed. Exposing it as an `asJson( strictMode = false )` argument would
  let callers choose.
- **First or last object.** `findJSONInText` returns the **first** balanced
  object. For the "model reconsiders itself" reply above, the first object is
  the answer the model *retracted* and the last is the one it settled on, so
  first-wins silently records the wrong verdict. Last-wins is the safer default
  for a self-correcting reply; first-wins is safer for a reply with a trailing
  example. No strong opinion, but the choice should be deliberate — right now
  neither is reachable because the call throws before getting there.

## Any logs, error output, etc?

The exception surfaces from `jsonDeserialize` with the whole reply as context.
A real example from our flight recorder (Bedrock, Claude Sonnet 4.6,
`max_tokens: 256`, temperature 0.1):

```
{"relation_type": "supersedes", "confidence": 0.72, "reasoning": "A restates B with newer framing, though this is closer to references."}

Wait, let me reconsider.

{"relation_type": "references", "confidence": 0.75, "reasoning": "A cites B; neither adds new claims to the other."}
```

Both objects are well-formed and the payload ends at a closing brace — nothing
was truncated, and `max_tokens` was not the constraint (successful calls at this
site peak at 47–91 output tokens against a 256 ceiling). The reply is simply not
*only* JSON, and that is enough to lose it.

Frequency, for scale: 30 calls over three days at one call site, peaking at
1.30% of a day's traffic. Small, but it is silent data loss on an otherwise
successful and fully-billed LLM call.

## Any other comments?

- Independent of #221, though they compound: #221 means a Bedrock caller cannot
  send `stop_sequences` (which would have suppressed the trailing object at the
  source), and this issue means the reply is then thrown away.
- We have worked around it locally by dropping to `.singleMessage()` and doing
  our own brace-matched extraction — i.e. re-implementing, at the call site,
  what `JSONExtractorTransformer` already does inside the module. That is the
  part that seems worth fixing upstream.
- Happy to raise a PR wiring `case "json"` through the transformer if you'd like
  it, once you've picked strict/lenient and first/last.

## What versions are you using?

**Operating System:** Amazon Linux 2023 (AWS Lambda, `java21` runtime, eu-west-2); also reproduced on macOS 15 (Darwin 25.6.0) under Docker
**Package Version:** bx-ai 3.3.0+13 on BoxLang 1.14.0+55 (same code on `development`); observed via Bedrock (`eu.anthropic.claude-sonnet-4-6`), but the line is identical in all five providers listed above

dougcain:
# Description

Fixes #222 — `returnFormat: "json"` (`.asJson()`) passed the model's whole reply to `jsonDeserialize()` after only stripping a leading markdown fence, so a reply containing valid JSON plus *any* surrounding prose threw and the call was lost. The module already ships `JSONExtractorTransformer` — a tolerant, string-aware brace/bracket scanner — but nothing wired it to the `asJson()` path.

Changes:

- `BaseService` gains a shared `extractJSONContent()` helper that routes through `JSONExtractorTransformer` (lenient mode); the five providers implementing the returnFormat switch (OpenAI, Claude, Gemini, Bedrock, Cohere) call it from `case "json"`. Everything extending `OpenAIService` (Groq, DeepSeek, Grok, Ollama, OpenRouter, Perplexity, MiniMax, Mistral, DockerModelRunner, HuggingFace, OpenAICompatible) inherits the fix.
- The transformer itself was hardened while wiring it in, after review caught real regressions in its fence handling:
  - fence extraction is now **parse-validated** — a ` ```json ` marker *inside a JSON string value* (e.g. `{"code":"...```json..."}`) no longer causes the reply to be mangled; every fence candidate is checked with `isJSON()` and the scan falls back to the original text when none parses;
  - the brace scan retries from each successive `{`/`[` until a candidate parses, so a prose brace before the payload (`Use {curly} like: {"a":1}`) no longer defeats extraction.
- Design decisions (flagged as open questions in the issue): **lenient** — an unusable reply returns `{}` rather than throwing, now accompanied by a warn-level entry in the `ai` log with a reply excerpt so the degraded path is diagnosable; **first valid object wins** (the transformer's existing semantics, now parse-validated).

Behaviour change to note: user code relying on the old throw-on-prose behaviour (e.g. `try/catch` around `.asJson().run()` as a retry trigger) will now receive `{}` + a warn log instead of an exception.

Deliberately out of scope (same bug class, separate path): `populateStructuredOutput()` / `SchemaBuilder` still call `jsonDeserialize` directly on model output for the `structuredOutput` return format — flagging rather than folding it in here to keep this diff reviewable.

**Testing:** 19 cases in `JSONExtractorTransformerTest` (clean object/array, leading/trailing prose, fenced blocks, fence-marker-inside-string-value, prose-brace-before-payload, scalar type preservation, garbage → `{}`), plus `MockServiceTest` cases exercising the full `aiChat( returnFormat: "json" )` path through the real provider switch (`MockService` extends `OpenAIService`; only HTTP transport is mocked). The prose and fence-in-string cases are red on `development`, green here. 92 tests across the affected suites, all passing.

Additionally verified inside a real consuming application's runtime: the discriminating cases (fence-marker-in-prose, invalid-first-fence) return `{}` on the stock module and the correct object with this fix, and the issue's original two-objects + "Wait, let me reconsider." reply parses. That application's integration suite shows zero regressions against the stock module.

## Jira/Github Issues

Fixes https://github.com/ortus-boxlang/bx-ai/issues/222
## Type of change

-   [x] Bug Fix
-   [ ] Improvement
-   [ ] New Feature

## Checklist

-   [x] I have made corresponding changes to the documentation


lmajano:
@dougcain please resolve conflicts

dougcain:
@lmajano conflicts resolved

Issue #227, opened 2026-08-02
Bedrock provider parity: broken ARN routing & credential chain, silent event/tool/usage degradations, and unexposed native capabilities (bearer auth, guardrails, Converse)


dougcain:
## Summary

A full review of `models/providers/BedrockService.bx` on `development` against the other providers (Claude, OpenAI, Gemini, Cohere, and the `BaseService` contract) turned up a set of gaps beyond the ones already filed (#221, #222, #226). Two of them prevent standard AWS deployments from running at all; several more are silent degradations where the API appears to work but quietly does less than every other provider; and a handful of cheap Bedrock-native capabilities (bearer-token auth, guardrails, request-level header knobs) have no route through the module.

Line references are against `development` (post-#224).

### Deployment blockers

1. **No IAM-role credential support, despite the docstring claiming it.** `loadAwsCredentialsFromEnvironment()` (L152-168) reads only `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_SESSION_TOKEN` / `AWS_REGION`. The class doc (L47) says "can use environment variables or IAM role" — but there is no IMDSv2, no ECS/EKS container-credentials endpoint (`AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`), no profile file, and no expiry/refresh for the session token it does accept. An app running on EC2/ECS/EKS under an instance role — the standard AWS deployment — cannot authenticate.

2. **Inference-profile ARN routing targets a route that does not exist.** `getBedrockPath()` emits `/application-inference-profile/{arn}/invoke` (L196-199, stream variant L221-224). The InvokeModel API only has `POST /model/{modelId}/invoke`, where `modelId` itself accepts inference-profile ARNs. On top of that the ARN is interpolated raw into the URL (L178, L210) while `AwsSignatureV4.uriEncodePath` percent-encodes `:` for the canonical request — a signature mismatch even if the route existed. So the documented `inferenceProfileArn` option (cited in `AiBaseRequest.bx` L112) can never have worked. The plain cross-region profile *id* form (`eu.anthropic.claude…`) works fine today.

### Silent degradations vs. the other providers

[...]

PR #233, opened 2026-08-02
feat: Bedrock provider parity — credential chain, bearer auth, ARN routing, guardrails, events/429/usage fixes

### LINKED ISSUE #258: Bedrock credential cache is per-instance but instances are never reused, so IMDS is re-hit on every call
## Problem

`BedrockService`'s AWS credential cache is **per-instance**, but service instances are **never
reused** — so on the primary code path the cache, its 5-minute refresh buffer, and its 60-second
negative cache all do nothing.

The chain:

- `aiChat()` → `aiService()` → `src/main/bx/bifs/aiService.bx:159`, which does
  `new "#targetClass#"().configure( arguments.options )`. A **fresh provider instance per call**;
  nothing caches or pools them.
- `BedrockService.init()` sets `variables.resolvedCredentials = {}` and
  `variables.credentialLockName = "bedrock-credentials-" & createUUID()` — both per-instance.

So every `aiChat()` against Bedrock constructs a new service, finds an empty
`resolvedCredentials`, and re-resolves credentials from scratch.

## Consequences

1. **Every request re-hits the container/IMDS metadata endpoint** on ECS/EKS/EC2 — the exact
   deployment the credential chain was written for. `isCredentialCacheValid()` is never `true` on
   a fresh instance, so the expiry-aware caching never engages.
2. **The negative cache never fires.** It was added so a non-AWS host would not pay an IMDS
   timeout on every request — but since `lastCredentialResolutionFailure` is also per-instance,
   every call pays the full timeout (2 × 1s in `loadImdsCredentials()`, plus the container probe).
   This directly contradicts the claim in #233's description, which I wrote; flagging it against
   myself.
3. **The exclusive lock is near-pointless in both directions.** @lmajano raised it as possible
   contention on #233 — but because `credentialLockName` embeds a per-instance UUID, two requests
   can never contend, so it protects nothing across requests. It only serializes threads sharing
   one explicitly-held instance.

Note the explicit/env credential path returns *before* the lock, so this is confined to the
auto-resolved container/IMDS path.

## Proposal

Move the credential cache (and the lock keyed to it) to **module scope**, shared across instances:

- Key the cache by what actually determines the credentials — region plus credential source
  (container endpoint vs IMDS role name) — not by instance identity.
- Derive the lock name from that same key, so concurrent refreshes for one key serialize while
  different keys proceed independently. This is what the existing exclusive lock was reaching for.
- Keep the existing expiry buffer and negative-cache semantics; they start working once the state
  outlives a single request.
- Storage options worth weighing: module settings/`application` scope vs the BoxLang cache. The
  cache carries credentials, so whatever is chosen must not be dumped in logs or serialized to a
  shared/persisted cache region.

Worth checking whether other providers have the same instance-per-call assumption baked in —
`AwsCredentialProvider.bx` (see the sibling issue) has its own chain and may share the flaw.

## Acceptance

- Two sequential `aiChat()` calls on an IMDS-backed host resolve credentials **once** (assert one
  metadata fetch, e.g. via a counter or an injected resolver).
- On a host with no metadata service, the second call does **not** pay the IMDS timeout, proving
  the negative cache spans instances.
- Concurrent calls for the same key trigger a single refresh, not one per thread.
- Explicit and environment credentials keep bypassing the whole mechanism, as today.

Found while triaging review feedback on #233.

dougcain:
# Description

Implements the mechanical/parity items of #227 (everything except real incremental streaming and the Converse migration, which the issue tracks as separate design discussions). Four commits: three feature chunks plus a review-fix pass.

**Fixed**

- **Interception events**: Bedrock announced `onAIRequest`/`onAIResponse` — names registered nowhere — so all Bedrock traffic was invisible to interceptors and the audit layer. Renamed to `onAIChatRequest`/`onAIChatResponse` with the same payload contract as the other providers (`chatRequest` key). `onAIError` now fires once per failure (previously twice, with contradictory `canRetry` values) with `canRetry` true for 429/5xx.
- **429 throttling** (Bedrock's most common error) now announces `onAIRateLimitHit` with `retry-after`, matching `BaseService`'s contract, on chat, stream and embeddings.
- **`super.configure()`** is now called, so `settings.defaultParams` and `settings.providers.Bedrock.{params,options}` merge like every other provider.
- **Inference-profile ARN routing**: the `/application-inference-profile/{arn}/invoke` path doesn't exist in the Bedrock API and the raw-interpolated ARN broke the SigV4 canonical path even where it could have matched. All requests now use `/model/{percent-encoded id-or-arn}/invoke(-with-response-stream)`, with the wire path and canonical request agreeing per AWS's non-S3 double-encode rule (verified against botocore's behaviour, and empirically that BoxLang's HTTP client transmits the encoded path untouched). This also fixes a latent signature mismatch for every model id containing `:`.
- **Claude responses** join all `type=="text"` content blocks instead of reading `content[1]` only — a leading `thinking` block or multiple text blocks no longer yields empty/truncated content. (`ClaudeService` still reads only the first block — noted as a follow-up, not touched here.)
- **Token usage** is no longer fabricated: Mistral/OpenAI-fallback paths compute real totals, all paths backfill from the `X-Amzn-Bedrock-*-Token-Count` response headers when the body lacks usage, and embeddings now report usage and announce `onAITokenCount`.
- **Array-shaped message content** (image parts, `tool_result` history) no longer breaks the Titan/Llama/Mistral transforms — text parts are flattened, non-text skipped.
- **Stream-path logging** (four unconditional info logs including a payload dump per streamed response) is now gated behind the existing log settings.

**Added**

- **AWS credential chain**: explicit struct → env vars → ECS/EKS container endpoint → EC2 IMDSv2, with expiry-aware caching (5-minute refresh buffer, still-valid cached creds survive a transient refresh failure, 60s negative cache so non-AWS hosts don't pay an IMDS timeout per request). Credentials resolve as an **atomic set** — an explicit access key defines the whole triple — so an explicit key pair can never inherit an unrelated `AWS_SESSION_TOKEN` from the environment and be signed as a mismatched triple. The cache is shared **across instances**, keyed by region + credential source, with the refresh lock derived from that key; `aiService()` builds a fresh provider per call, so an instance-scoped cache was written once and never read and the negative cache could never fire ([#258](https://github.com/ortus-boxlang/bx-ai/issues/258)). The class docblock previously claimed IAM-role support that didn't exist; now it does, and the docs match.
- **Bedrock API-key (bearer) auth**: an explicit `bearerToken` config key or the `AWS_BEARER_TOKEN_BEDROCK` environment variable selects `Authorization: Bearer`, skipping SigV4 entirely — no AWS credentials needed. Explicit/env AWS credentials always take precedence, and bearer auth is never inferred from a generic `apiKey`, so a module-wide `settings.apiKey` (which `aiService()` injects into every provider) can neither disable signing nor be mistaken for a Bedrock key. Bedrock also joins `ProviderCapabilitiesTest`.
- **Guardrails**: `providerOptions.guardrailIdentifier/guardrailVersion/guardrailTrace` → the corresponding `X-Amzn-Bedrock-*` request headers (SigV4-signed); `INTERVENED` responses are logged to the `ai` log.
- **Generic header passthrough**: `providerOptions.bedrockHeaders` struct or any `x-amzn-bedrock-*` providerOption becomes a signed request header (service tier, latency-optimized inference, request metadata, trace).
- **`baseURL` endpoint override** (config or `BOXLANG_MODULES_BXAI_BEDROCK_BASEURL`), including path-prefixed proxy shapes — enables VPC interface endpoints, FIPS hosts, and LocalStack.
- **Embeddings shapes**: `cohere.embed-*` (correct `texts`/`input_type` body, batched 96 per call, order-preserving) and Titan v2 (`dimensions`/`normalize`) alongside the existing Titan v1 default.
- **Loud capability errors**: requesting tools or schema-typed structured output for a model family whose transform genuinely drops them (titan/llama/mistral/openai/cohere) now throws `UnsupportedProviderCapability` instead of silently sending nothing.

**Fully backwards-compatible.** `configure( "string" )` keeps its long-standing Bedrock meaning of setting `modelId` — Bedrock deliberately deviates from `BaseService`'s "string means apiKey" contract because a model id is the more useful single-string argument here, and that intent is now documented in the code. Bearer auth is opt-in via the new `bearerToken` key or AWS's own env var, so nothing about existing configuration changes behaviour.

**Testing:** 30+ new cases across `BedrockServiceTest` and `ProviderCapabilitiesTest` — path/ARN encoding (including a wire-vs-canonical `%3A`/`%253A` agreement test against a new `AwsSignatureV4.computeCanonicalPath()` seam), bearer-vs-SigV4 precedence (explicit AWS creds beat an explicit `bearerToken`; a bare `apiKey` never selects bearer auth), baseURL with and without path prefix, header/guardrail assembly, capability-gate throw/allow, content flattening, embeddings shape routing and request builders, credential-chain resolution and expiry-buffer logic, settings merge, multi-block content join. 63/64 green locally (the one failure is a pre-existing live-network test). Container/IMDS endpoints are unreachable from tests and are noted as such in code.

## Jira/Github Issues

Refs https://github.com/ortus-boxlang/bx-ai/issues/227 (implements all items except streaming (10) and Converse (17), which remain open for design discussion)

Closes [#258](https://github.com/ortus-boxlang/bx-ai/issues/258) (credential cache was per-instance while instances are never reused, so IMDS was re-hit on every call and the negative cache never fired — a defect in this PR's own credential chain, hence fixed here rather than deferred).

Partially addresses [#259](https://github.com/ortus-boxlang/bx-ai/issues/259): the `chr()` → `char()` convention fix in `AwsCredentialProvider.bx` is included. Consolidating that file's separate AWS credential chain with this one is **not** — it changes `OpenSearchVectorMemory`'s authentication behaviour and doesn't belong in a Bedrock parity PR, so #259 stays open for it.

### Review follow-ups (post-review commits)

- Four response-header reads normalized through `firstHeaderValue()`: the guardrail-action header (an array-shaped value made `findNoCase()` resolve to `arrayFindNoCase` and silently miss the intervention) and `retryAfter` in all three 429 handlers (listeners got an array, so `val( retryAfter )` yielded `0` instead of the real backoff).
- `buildProviderOptionHeaders()`'s docblock listed bare suffixes as if they were valid keys, which had misled a test into asserting header names Bedrock would ignore. Docblock corrected, test uses full `X-Amzn-Bedrock-*` names, and a new test pins that nothing prefixes `bedrockHeaders` for the caller.
- `guardrailTrace` deliberately keeps mapping to `X-Amzn-Bedrock-Trace`, not `X-Amzn-Bedrock-GuardrailTrace`. AWS's InvokeModel reference is self-inconsistent — its Request Syntax and URI Request Parameters document only the former (`ENABLED / DISABLED / ENABLED_FULL`), and it is what produces the response body's `amazon-bedrock-trace.guardrails` block, while the latter appears only in the "Use a guardrail" example as a parameter neither authoritative section defines. Both are reachable via the generic passthrough; the reasoning is recorded in the docblock.

## Type of change

-   [x] Bug Fix
-   [x] Improvement
-   [x] New Feature

## Checklist

-   [x] I have made corresponding changes to the documentation


src/main/bx/models/providers/BedrockService.bx
```diff
@@ -174,75 +252,417 @@ class extends = "BaseService" implements="IAiChatService, IAiEmbeddingsService"
...
-	 * @modelIdOrArn The model ID or inference profile ARN to invoke
+	 * @return Struct { accessKeyId, secretAccessKey, sessionToken, expiration } or null if the
+	 *         endpoint isn't configured or the request fails
+	 */
+	private any function loadContainerCredentials(){
+		var relativeUri = getSystemSetting( "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI", "" );
+		var fullUri      = getSystemSetting( "AWS_CONTAINER_CREDENTIALS_FULL_URI", "" );
+
+		if ( !len( relativeUri ) && !len( fullUri ) ) {
+			return javacast( "null", "" );
```
lmajano: Please change to just `null` in BoxLang

src/main/bx/models/providers/BedrockService.bx
```diff
@@ -174,75 +252,417 @@ class extends = "BaseService" implements="IAiChatService, IAiEmbeddingsService"
...
 	 *
-	 * @return The full streaming endpoint URL
+	 * @return Struct { accessKeyId, secretAccessKey, sessionToken, expiration } or null if IMDS
+	 *         is unreachable or no role is attached
 	 */
-	private string function getBedrockStreamEndpoint( required string modelIdOrArn ){
-		return "https://bedrock-runtime.#variables.region#.amazonaws.com#getBedrockStreamPath( arguments.modelIdOrArn )#";
+	private any function loadImdsCredentials(){
+		try {
+			var tokenResult = http( url = "http://169.254.169.254/latest/api/token" )
```
lmajano, line 832: why the IP address?

src/main/bx/models/providers/BedrockService.bx
dougcain: Done in 8da4af2 — all 9 `javacast( "null", "" )` occurrences this branch added to `BedrockService.bx` (the credential-chain returns plus the two `lastCredentialResolutionFailure` assignments) are now the bare `null` literal.

Left alone: 3 pre-existing occurrences elsewhere in this file (in `transformStreamChunk`/`parseStreamChunk`, incl. a `"finish_reason" : ... : javacast( "null", "" )` ternary) and ~13 other files across `src/main/bx` still use the old form — say the word and I'll sweep them in a separate PR rather than widen this diff.

Also merged the latest `development` (the #236 `gatewayRegistry()` changelog entry superseded the earlier dotted-name wording) — back to MERGEABLE. Bedrock + capabilities suites: 64/65 green, the single failure being the live-AWS `testRealClaudeCall`, which needs real credentials.

lmajano:
@dougcain ok. Please rebase with development and retest your feature. I am ready for release

dougcain:
@lmajano Merged with `development` and retested — head is now `1ad270e`.

The merge was clean, no conflicts. Your earlier `null`-literal note was already addressed in
`8da4af2`.

**Please read the red checks before judging this one.** Both failing jobs are environmental, not
code:

- **`Tests bx-ai`** — this PR is from a fork, so the run gets no repo secrets. 153 of the 154
  failures are missing-API-key (`aiEmbedTest`, the vector-memory integrations, `ClaudeTest`,
  `GeminiTest`, …), and `SummaryMemoryTest`'s two are the same cause — its summarization does a
  live LLM call, and it passes on `development` where secrets exist. **Zero Bedrock failures.**
- **`Checks Source Code Formatting`** — shows 24h runtime because it never got a runner
  ("exceeded the maximum execution time while awaiting a runner"). Not a spotless violation;
  `spotlessCheck` passes locally.

## What changed since your review

Three things the merge itself surfaced, all Bedrock-only:

- **Bedrock reported reasoning only when streaming.** `development` now normalizes reasoning onto
  `message.reasoning` / `delta.reasoning` and patched Bedrock's *stream* path — but Bedrock
  overrides `chat()` wholesale, so it never reaches `BaseService.sendChatRequest()` where
  `normalizeReasoningMessage()` runs, and its Claude transform filtered content to `type == "text"`
  and dropped `thinking` blocks outright. Claude-on-Bedrock extended thinking now surfaces as
  `message.reasoning`, and DeepSeek-on-Bedrock's native `reasoning_content` is normalized too.
  Nothing caught this: `ReasoningNormalizationTest` is MockService-only with no Bedrock coverage,
  so I added three cases.
- **`beforeLLMCall` could not mutate the request body on the streaming path.**
  `sendBedrockStreamRequest()` serialized `dataPacket` *before* firing the hook, so anything
  middleware changed was signed and sent as the pre-hook body. The sync path already ordered these
  correctly.
- **Explicit AWS credentials could be signed with an unrelated session token** — the one I'd most
  want a second opinion on. `init()` seeds the instance from the environment, so a later
  `configure({ awsAccessKeyId, awsSecretAccessKey })` carrying no token of its own replaced only
  the key pair and kept `AWS_SESSION_TOKEN` from the environment. SigV4 then signed a mismatched
  triple and AWS rejected it. Credentials now resolve as an atomic set: an explicit access key
  defines the whole set, so the secret and session token come from that same source or not at all.
  Partial overrides of secret/token alone keep their previous behaviour.

Also tightened test hygiene: the two live-AWS tests now use `assumeTrue` so they report **skipped**
rather than falsely **passed**, and the live smoke test aborts on environment facts (unknown key,
`AccessDenied`) while deliberately still failing the build on `InvalidSignatureException` and
`ValidationException` — those are what a real SigV4 or request-transform regression looks like.
The long-standing "1 pre-existing failure" is now a clean skip.

## Local results

145 tests, 0 failures, 1 skip:

| Suite | Result |
|---|---|
| `BedrockServiceTest` | 54 — 53 passed, 1 skipped (live AWS call, no valid credentials) |
| `ProviderCapabilitiesTest` | 19 |
| `ReasoningNormalizationTest` | 8 |
| `RunControlMiddlewareTest` | 12 |
| `PromptSecurityTest` / `RedactTest` / `FenceTest` | 15 / 18 / 9 |
| `OutputGuardMiddlewareTest` | 9 |

`spotlessCheck` and `git diff --check` pass.

One note if you want to align it: `development`'s Claude sync fix selects the **first** text block,
while Bedrock joins **all** of them. Mine is arguably the more correct behaviour, but the two
providers now differ — happy to change whichever way you prefer.

#249 is stacked on this and rebased on top of the same head, so it's ready the moment this lands.

src/main/bx/models/providers/BedrockService.bx
```diff
@@ -1184,6 +1796,107 @@ class extends = "BaseService" implements="IAiChatService, IAiEmbeddingsService"
...
+	 * @httpResult The raw HTTP result (for response headers)
+	 * @response   The parsed JSON response body
+	 */
+	private void function checkGuardrailIntervention( required struct httpResult, required struct response ){
+		var action          = "";
+		var responseHeaders = arguments.httpResult.responseHeader ?: {};
+		if ( isStruct( responseHeaders ) ) {
+			for ( var headerName in responseHeaders ) {
+				if ( headerName.reFindNoCase( "^x-amzn-bedrock-guardrailaction$" ) ) {
+					action = responseHeaders[ headerName ];
```
lmajano: **Correctness: normalize this header read through `firstHeaderValue()`**

Every other header read added in this PR goes through `firstHeaderValue()` to guard against an array-shaped multi-value header (see its own docblock a few lines up, and the token-count reads at lines 2070–2071). This one reads the raw header value directly:

```boxlang
action = responseHeaders[ headerName ];
```

If the HTTP client ever returns `x-amzn-bedrock-guardrailaction` array-shaped, `action` becomes an array, and `action.findNoCase( "INTERVENED" )` two lines below throws (`findNoCase` is string-only) — since `checkGuardrailIntervention()` is called unguarded from both `sendBedrockRequest()` and `sendBedrockStreamRequest()`, that turns an otherwise-successful 200 response into a hard failure.

Suggested fix:
```boxlang
action = firstHeaderValue( responseHeaders[ headerName ] );
```

---

src/main/bx/models/providers/BedrockService.bx
```diff
@@ -1287,6 +2019,32 @@ class extends = "BaseService" implements="IAiChatService, IAiEmbeddingsService"
...
+				var responseHeaders = httpResult.responseHeader ?: {};
+				boxAnnounce(
+					"onAIRateLimitHit",
+					{
+						provider    : this,
+						operation   : "chat",
+						statusCode  : statusCode,
+						errorData   : errorData,
+						chatRequest : arguments.aiRequest,
+						retryAfter  : responseHeaders.keyExists( "retry-after" ) ? responseHeaders[ "retry-after" ] : ""
```
lmajano: **Correctness: `retryAfter` isn't normalized via `firstHeaderValue()` either**

Same pattern as the guardrail-action header above — this reads `retry-after` directly:

```boxlang
retryAfter  : responseHeaders.keyExists( "retry-after" ) ? responseHeaders[ "retry-after" ] : ""
```

If the client returns `Retry-After` array-shaped, `onAIRateLimitHit` listeners get an array instead of a string/number in `retryAfter`. A rate-limit-aware retry middleware doing `val( retryAfter )` or parsing it as a duration would get `0` or throw instead of the real backoff — undermining exactly the 429-handling this PR adds.

Same issue repeats at line 2287 (stream) and line 3100 (embeddings) — worth fixing all three together.

Suggested fix (matches the existing `firstHeaderValue( httpResult.responseHeader[ "..." ] ?: "" )` pattern used for the token-count headers at lines 2070–2071):
```boxlang
retryAfter  : firstHeaderValue( responseHeaders[ "retry-after" ] ?: "" )
```

---

src/main/bx/models/providers/BedrockService.bx
```diff
@@ -1459,6 +2264,31 @@ class extends = "BaseService" implements="IAiChatService, IAiEmbeddingsService"
...
+				var responseHeaders = httpResult.responseHeader ?: {};
+				boxAnnounce(
+					"onAIRateLimitHit",
+					{
+						provider    : this,
+						operation   : "stream",
+						statusCode  : statusCode,
+						errorData   : errorData,
+						chatRequest : arguments.aiRequest,
+						retryAfter  : responseHeaders.keyExists( "retry-after" ) ? responseHeaders[ "retry-after" ] : ""
```
lmajano: Same `retryAfter` normalization gap as the chat-path 429 handler (see comment on line 2043) — this is the stream-path copy of the same code:

```boxlang
retryAfter  : responseHeaders.keyExists( "retry-after" ) ? responseHeaders[ "retry-after" ] : ""
```

Suggested fix:
```boxlang
retryAfter  : firstHeaderValue( responseHeaders[ "retry-after" ] ?: "" )
```

---

src/main/bx/models/providers/BedrockService.bx
```diff
@@ -2028,6 +3076,32 @@ class extends = "BaseService" implements="IAiChatService, IAiEmbeddingsService"
...
+				var responseHeaders = httpResult.responseHeader ?: {};
+				boxAnnounce(
+					"onAIRateLimitHit",
+					{
+						provider         : this,
+						operation        : "embeddings",
+						statusCode       : statusCode,
+						errorData        : errorData,
+						embeddingRequest : arguments.embeddingRequest,
+						retryAfter       : responseHeaders.keyExists( "retry-after" ) ? responseHeaders[ "retry-after" ] : ""
```
lmajano: Same `retryAfter` normalization gap as the chat/stream 429 handlers (see comments on lines 2043 and 2287) — this is the embeddings-path copy:

```boxlang
retryAfter       : responseHeaders.keyExists( "retry-after" ) ? responseHeaders[ "retry-after" ] : ""
```

Suggested fix:
```boxlang
retryAfter       : firstHeaderValue( responseHeaders[ "retry-after" ] ?: "" )
```

---

src/main/bx/models/providers/BedrockService.bx
```diff
@@ -175,75 +269,476 @@ class extends = "BaseService" implements="IAiChatService, IAiEmbeddingsService"
...
+			if ( static.DEFAULT_CHAT_PARAMS.keyExists( key ) && static.DEFAULT_CHAT_PARAMS[ key ] == variables.params[ key ] ) {
+				continue;
+			}
+			configured[ key ] = variables.params[ key ];
+		}
+
+		return configured;
+	}
+
+	public struct function resolveAwsCredentials(){
```
lmajano, line 414: **Minor/efficiency (not blocking): exclusive lock on every call, even warm-cache reads**

`resolveAwsCredentials()` takes an exclusive named lock on every chat/stream/embeddings call when using auto-resolved (container/IMDS) credentials — including the common case where `isCredentialCacheValid()` is `true` and it's just returning the cached struct. Under concurrent Bedrock traffic on EC2/ECS (the exact deployment this PR targets), every request serializes through one lock for what's normally a read-only check.

Not a correctness bug (lock hold time is CPU-only, no I/O) — just worth considering a read/check-then-exclusive-refresh pattern (e.g. an unlocked fast-path check of `isCredentialCacheValid()`, only taking the exclusive lock when an actual refresh is needed) if this becomes a contention point in practice. Feel free to leave as-is for this PR and revisit later.

---

lmajano (commented):
Reviewed the diff (mostly `BedrockService.bx` and `AwsSignatureV4.bx`). Overall solid — the credential-chain, ARN routing, event renaming, and content-flattening fixes all check out against their call sites. Left inline comments with concrete one-line fixes for two related correctness gaps (both follow the same `firstHeaderValue()` normalization pattern already used elsewhere in this file), plus one lower-priority efficiency note:

1. `checkGuardrailIntervention()` (line 1882) reads the guardrail-action header directly instead of through `firstHeaderValue()` — an array-shaped header would throw inside an unguarded call path.
2. `retryAfter` in the `onAIRateLimitHit` payload has the same unnormalized read in all three 429 handlers (chat @2043, stream @2287, embeddings @3100).
3. (non-blocking) `resolveAwsCredentials()` takes an exclusive lock even on the warm-cache read path (line 352) — worth a look if it becomes a contention point.

Separately, unrelated to this diff: `src/main/bx/models/util/AwsCredentialProvider.bx` (not touched by this PR) uses `chr()` instead of the project's documented `char()` convention — flagging for a future pass, not blocking this PR.

---

src/main/bx/models/providers/BedrockService.bx
dougcain: Fixed in dd2caac — now `action = firstHeaderValue( responseHeaders[ headerName ] );`.

One correction to the mechanism, because it changes the severity rather than the fix: arrays **do**
have a `findNoCase` member function, so `action.findNoCase( "INTERVENED" )` on an array resolves to
`arrayFindNoCase` against the header *values* and returns `0` — it doesn't throw. So the real
failure is a **silently missed guardrail intervention**, not a hard failure on an otherwise-good
200. The throw you describe would happen one line later, interpolating the array into
`"...(action=#action#)"`, and only in the case where it *did* match.

Silent is arguably the worse of the two for something whose entire job is to report that a guardrail
fired, so the fix stands as suggested.

No test added: `checkGuardrailIntervention()` is private and runs inside `sendBedrockRequest()`,
which a canned `wrapLLMCall` response bypasses entirely, so there's no offline seam to assert
through — I'd have to widen the public API to test a one-line normalization onto an existing helper.
Flagging that rather than adding a test that proves nothing.

src/main/bx/models/providers/BedrockService.bx
dougcain: Agreed and fixed in dd2caac — all three sites (chat, stream, embeddings) now read:

```boxlang
retryAfter  : firstHeaderValue( responseHeaders[ "retry-after" ] ?: "" )
```

Your `val( retryAfter )` point is the one that made this worth doing over the guardrail one: the
whole reason the 429 handling exists is so a retry middleware can back off correctly, and handing it
an array meant `val()` → `0` → immediate retry into the same throttle.

Same testability caveat as the guardrail comment: these live in private paths inside
`sendBedrockRequest()` / `sendBedrockStreamRequest()` / `embeddings()`, which a canned `wrapLLMCall`
response bypasses, so there's no offline seam. Not adding a test rather than inventing public API
for a one-liner onto an existing helper.

src/main/bx/models/providers/BedrockService.bx
dougcain: Fixed in dd2caac along with the chat-path and embeddings copies — all three now go through
`firstHeaderValue()`. See the reply on line 2043 for the detail.

src/main/bx/models/providers/BedrockService.bx
dougcain: Fixed in dd2caac along with the chat-path and embeddings copies — all three now go through
`firstHeaderValue()`. See the reply on line 2043 for the detail.

src/main/bx/models/providers/BedrockService.bx
dougcain: Thanks for this one — chasing it turned up something worse than the contention you flagged, and in the
opposite direction, so it's **fixed in this PR** rather than deferred.

The lock name was per-instance (`"bedrock-credentials-" & createUUID()`), and `aiService.bx:159` builds
a fresh provider on every call, so two requests could never contend — the lock protected nothing across
requests. But the same fact meant `variables.resolvedCredentials` was empty every time: the expiry
cache never engaged, so every request re-hit the container/IMDS endpoint, and the 60s negative cache
never fired either, so a non-AWS host paid the full IMDS timeout on *every* call. That last part
contradicted a claim in this PR's own description; that one's on me.

Fixed by moving the cache to module scope keyed by region + credential source, with the lock derived
from that key — which is what your comment was reaching for. Then a follow-up review caught that this
introduced two new problems, both also fixed here:

- The shared lock became **JVM-wide** (BoxLang's lock registry is static) and was held across ~5s of
  container+IMDS HTTP plus an untimed token-file read. Past the 10s timeout it throws `type="lock"`,
  which `RetryMiddleware`'s `nonRetryableTypes` doesn't cover, so it would sleep and re-queue onto the
  same congested lock. Reads now take a `readonly` lock, resolution happens **outside** any lock, and
  only the write-back is exclusive.
- The shared negative cache could turn one cold-start miss into a runtime-wide outage: it served empty
  credentials — a guaranteed 403 — to every request for the full 60s, where per-instance state had let
  the next request simply retry. The window is now 60s only while usable credentials remain (pure
  refresh suppression) and 5s when there's nothing to serve.

Also dropped `region` from the cache key (the metadata service returns the same credentials regardless,
so it bounded nothing while letting unvalidated caller-supplied region strings mint unbounded entries
each holding a live secret), and `resolveAwsCredentials()` now returns a copy rather than the live
shared struct.

src/main/bx/models/providers/BedrockService.bx
```diff
@@ -1184,6 +1796,107 @@ class extends = "BaseService" implements="IAiChatService, IAiEmbeddingsService"
...
 	}
 
+	/**
+	 * Build extra request headers driven by providerOptions:
+	 *   - Generic passthrough: any providerOptions key already shaped like a Bedrock header
+	 *     (matches `x-amzn-bedrock-*`, case-insensitive) is forwarded verbatim. Covers ad-hoc
+	 *     headers like PerformanceConfig-Latency, Service-Tier, Request-Metadata, Trace.
+	 *   - Struct shorthand: `providerOptions.bedrockHeaders = { "X-Amzn-Bedrock-...": "value" }`.
+	 *   - First-class Guardrails: `guardrailIdentifier` / `guardrailVersion` / `guardrailTrace`
+	 *     map to X-Amzn-Bedrock-GuardrailIdentifier / -GuardrailVersion / -Trace.
```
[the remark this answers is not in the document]
dougcain: Checked this against the [InvokeModel API reference](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html) — **keeping `X-Amzn-Bedrock-Trace`**, but you've correctly spotted that AWS's own docs contradict themselves, so I've recorded the reasoning in the docblock rather than leaving it to look like an oversight.

What the page actually says:

- **Request Syntax** lists `X-Amzn-Bedrock-Trace: {{trace}}` alongside `-GuardrailIdentifier` and
  `-GuardrailVersion`. No `GuardrailTrace`.
- **URI Request Parameters** documents `trace` with `Valid Values: ENABLED | DISABLED | ENABLED_FULL`.
  There is no `guardrailTrace` parameter at all.
- The **"Use a guardrail" example** shows `X-Amzn-Bedrock-GuardrailTrace: guardrailTrace` *and*
  `X-Amzn-Bedrock-Trace: trace` as two separate headers.

So `GuardrailTrace` appears only in an example, as a parameter neither authoritative section defines,
while `X-Amzn-Bedrock-Trace` is formally specified — and the page's own guardrail *response* example
confirms it's what yields the `amazon-bedrock-trace.guardrails` block in the body. Sending the
documented header is the safer default.

Both remain reachable regardless: the generic `x-amzn-bedrock-*` passthrough forwards
`X-Amzn-Bedrock-GuardrailTrace` verbatim if a caller wants the example's variant.

Fixed in dd2caac: the docblock no longer implies `guardrailTrace` maps to a guardrail-specific
header, and now states the inconsistency, the values, and why we chose this one, with the doc link.
The test assertion stays on `X-Amzn-Bedrock-Trace` since that's the intended behaviour.

src/test/java/ortus/boxlang/ai/providers/BedrockServiceTest.java
```diff
@@ -684,6 +740,1390 @@ public void testProviderOptions() {
...
+					{ awsAccessKeyId: "%s", awsSecretAccessKey: "%s", region: "%s" }
+				)
+				headers = service.buildProviderOptionHeaders( {
+					bedrockHeaders: {
+						"Service-Tier": "flex",
+						"Request-Metadata": "project=demo"
+					}
+				} )
+				serviceTier = headers[ "Service-Tier" ]
+				requestMeta = headers[ "Request-Metadata" ]
```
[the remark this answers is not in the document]
dougcain: Good catch — this was a real hole in the test, fixed in dd2caac.

You're right that the test was asserting behaviour for header names Bedrock won't recognize.
`bedrockHeaders` is appended **verbatim** (`extraHeaders.append( bedrockHeaders, true )`), so
`"Service-Tier"` would have gone on the wire as a literal `Service-Tier` header and been ignored —
and the test passed anyway, because it only checked that the key round-tripped through the struct.

Root cause was the docblock, which listed bare *suffixes* — "PerformanceConfig-Latency, Service-Tier,
Request-Metadata, Trace" — as if they were valid keys, when the generic passthrough regex anchors on
`^x-amzn-bedrock-`. So the test was faithfully implementing a misleading doc.

Changes:

- `testBedrockHeadersStructShorthand` now uses full `X-Amzn-Bedrock-Service-Tier` /
  `X-Amzn-Bedrock-Request-Metadata` names.
- Added `testBedrockHeadersAreNotPrefixed`, which pins that a bare suffix is forwarded as-is and
  **not** prefixed for the caller — so nobody "fixes" this later by silently adding the prefix and
  changing what goes on the wire.
- Docblock now states the full-name requirement explicitly for both the passthrough and the struct
  shorthand.

src/main/bx/models/providers/BedrockService.bx
```diff
@@ -1184,6 +1796,107 @@ class extends = "BaseService" implements="IAiChatService, IAiEmbeddingsService"
...
+	 */
+	private void function checkGuardrailIntervention( required struct httpResult, required struct response ){
+		var action          = "";
+		var responseHeaders = arguments.httpResult.responseHeader ?: {};
+		if ( isStruct( responseHeaders ) ) {
+			for ( var headerName in responseHeaders ) {
+				if ( headerName.reFindNoCase( "^x-amzn-bedrock-guardrailaction$" ) ) {
+					action = responseHeaders[ headerName ];
+					break;
+				}
```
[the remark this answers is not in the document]
dougcain: Same finding as the review comment on line 1882 — fixed in dd2caac via `firstHeaderValue()`. One
note on the mechanism: arrays do have a `findNoCase` member function, so this resolves to
`arrayFindNoCase` and returns `0` rather than throwing, meaning the failure is a silently missed
guardrail intervention. Detail in the [thread on 1882](https://github.com/ortus-boxlang/bx-ai/pull/233#discussion_r3806259420).

src/main/bx/models/providers/BedrockService.bx
dougcain: Sorry — this one slipped through two review rounds unanswered.

`169.254.169.254` is the fixed link-local address AWS mandates for the EC2 Instance Metadata Service.
It isn't a magic number we chose or a host that could be configured: it's the same address on every
EC2 instance in every region, it's non-routable (the 169.254.0.0/16 link-local block, so it never
leaves the instance), and it's how every AWS SDK reaches IMDS. There's no DNS name for it.

The two calls are the IMDSv2 handshake: `PUT /latest/api/token` for a session token, then
`GET /latest/meta-data/iam/security-credentials/` with that token to discover the attached
instance-profile role. IMDSv1 allowed the second call without the token; v2 requires it, which is
what makes it resistant to SSRF.

Both use `.timeout( 1 )` and `.throwOnError( false )` so a non-EC2 host fails fast and falls through
the chain rather than hanging. Docblock links
[the AWS IMDSv2 reference](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-metadata-v2-how-it-works.html)
a few lines above.

Worth noting: the fall-through is currently more expensive than intended, because the negative cache
meant to absorb that timeout is per-instance and instances aren't reused — filed as #258.

Happy to lift the address into a named constant (e.g. `static.IMDS_ENDPOINT`) if you'd rather it not
read as a bare literal.

dougcain:
@lmajano All review comments addressed in dd2caac (head now `dd2caac`). Replied inline on each
thread; summary here.

**Fixed — the four unnormalized header reads.** Both of your correctness findings, plus Copilot's
duplicate of the first:

- `checkGuardrailIntervention()` now reads through `firstHeaderValue()`.
- `retryAfter` normalized in all three 429 handlers (chat 2043, stream 2287, embeddings 3100).

One correction on the guardrail one, since it changes severity rather than the fix: arrays *do* have
a `findNoCase` member function, so that call resolves to `arrayFindNoCase` and returns `0` instead of
throwing — the real failure is a **silently missed** intervention, not a hard failure on a good 200.
Silent is arguably worse for something whose only job is to report that a guardrail fired.

I did **not** add tests for these four. All of them sit in private paths inside
`sendBedrockRequest()` / `sendBedrockStreamRequest()` / `embeddings()`, which a canned `wrapLLMCall`
response bypasses entirely — so testing them means widening the public API for a one-line
normalization onto an existing helper. Flagging that explicitly rather than adding a test that only
proves the struct round-trips.

**Copilot's guardrail-trace finding — investigated, not changing the header.** AWS's InvokeModel
reference contradicts itself: the Request Syntax and URI Request Parameters document only
`X-Amzn-Bedrock-Trace` (`ENABLED | DISABLED | ENABLED_FULL`) and no `guardrailTrace` parameter at
all, while the "Use a guardrail" *example* shows `X-Amzn-Bedrock-GuardrailTrace` alongside it. We
send the formally documented one, which is also what produces the response body's
`amazon-bedrock-trace.guardrails` block; the example's variant is reachable via the generic
passthrough. Reasoning and doc link now recorded in the docblock so it doesn't read as an oversight.

**Copilot's test finding was a genuine hole.** `testBedrockHeadersStructShorthand` asserted bare
`Service-Tier` / `Request-Metadata` keys, which `bedrockHeaders` forwards verbatim — so those would
have hit the wire as unrecognized header names while the test passed. Root cause was the docblock
listing bare suffixes as if they were valid keys. Test now uses full names, plus a new test pinning
that nothing prefixes them for the caller, and the docblock states the rule for both entry points.

**Your lock note led somewhere worse — filed as #258.** The mechanism is right but the blast radius
is inverted: `credentialLockName` embeds a per-instance `createUUID()` and `aiService.bx:159`
constructs a fresh provider on every call, so two requests can never contend. The same fact means
`resolvedCredentials` is empty every time — so the expiry cache never engages and IMDS is re-hit per
request, and the 60s negative cache never fires either, so a non-AWS host pays the full IMDS timeout
on every call. That last part contradicts a claim in this PR's own description; that one's on me.
Left the lock alone here as you suggested, since changing it in isolation just moves furniture.

**Your `chr()` note — filed as #259**, together with something more interesting: that file isn't
untouched leftovers, it's live. `OpenSearchVectorMemory.bx:701` uses `AwsCredentialProvider.bx`, so
the module has two independent AWS credential chains that have drifted — meaning the EKS Pod Identity
fix in this PR landed in only one of them.

Also answered your Aug 4 question about `169.254.169.254` — sorry, that sat unanswered through two
rounds.

**Tests:** `BedrockServiceTest` 55 — 54 passed, 1 skipped (live AWS call, no valid credentials).
`spotlessCheck` passes. #249 has been rolled forward onto this head and re-verified (117 tests, 0
failures).

src/main/bx/models/providers/BedrockService.bx
dougcain: Follow-up: this is now **fixed in this PR** rather than deferred — 2133508, and #258 is closed.

On reflection it belonged here: the broken cache is a defect in the credential chain this PR
introduces, not a pre-existing issue, and the PR description advertised a negative cache that could
never fire. Shipping that as-is would have meant releasing the claim without the behaviour.

What changed:

- Auto-resolved credentials moved to `static.CREDENTIAL_CACHE`, shared across instances, keyed by
  `credentialCacheKey()` — region plus credential source (container relative URI / container full
  URI / `ec2-imds`).
- The named lock is now derived from that key instead of a per-instance UUID, so concurrent refreshes
  for one source collapse into a single metadata round-trip while different sources don't serialize —
  which is what your comment was pointing at in the first place.
- `fallbackToCachedOrEmptyCredentials()` takes the cached struct as a parameter rather than reading
  instance state. Explicit/env credentials still return before the lock.

`testCredentialCacheIsSharedAcrossInstances` covers it, and I checked it has teeth: re-keying the
cache per-instance makes it fail. Its timing bound is generous on purpose so it can't flake either
off-AWS or on.

Your `chr()` note is also fixed here (`char()` in the two spots). The remaining half of #259 —
consolidating that file's separate chain with `BedrockService`'s — stays open; it changes
`OpenSearchVectorMemory` auth and shouldn't ride along in this PR. Worth flagging that #258 widened
that gap: OpenSearch vector memory now lacks both Pod Identity support and credential caching, the
two things we just fixed here.

dougcain:
@lmajano Follow-up to the last round: the two issues I'd split out are now **in this PR**, head
`2133508`.

On reflection #258 belonged here rather than deferred. The broken credential cache isn't a
pre-existing issue — it's a defect in the chain this PR introduces, and the PR description advertised
a 60-second negative cache that could never fire. Releasing that would have shipped the claim without
the behaviour.

**#258 — credential cache now shared across instances (closed).** `resolvedCredentials`, the
negative-cache timestamp and the lock name were all per-instance while `aiService.bx` builds a fresh
provider per call, so the cache was written once and never read: every request re-hit
container/IMDS, and a host with no metadata service paid the full timeout on *every* call instead of
once a minute. Auto-resolved credentials now live in a cache keyed by region + credential source,
with the lock derived from that key so concurrent refreshes for one source collapse into a single
round-trip while different sources don't serialize — which is what your original note was pointing
at. Explicit/env credentials still bypass it entirely.

**#259 — `chr()` → `char()` done; the rest stays open.** The convention fix is in. Consolidating
`AwsCredentialProvider.bx`'s separate credential chain with this one is deliberately *not* here — it
changes `OpenSearchVectorMemory`'s auth behaviour and shouldn't ride along in a parity PR you're
queuing for release. Worth knowing that #258 widened that gap: OpenSearch vector memory now lacks
both Pod Identity support and credential caching, the two things fixed on the Bedrock side.

**One new public method:** `credentialCacheKey()`, consistent with `resolveAwsCredentials()`,
`isCredentialCacheValid()` and `buildProviderOptionHeaders()` already being public here. It returns a
hash of region + source — no credential material.

**Tests.** `BedrockServiceTest` 56 — 55 passed, 1 skipped (live AWS call, no valid credentials);
`AwsCredentialProviderTest` 8 green after the `char()` change; `OutputGuardMiddlewareTest` 12,
`PromptSecurityTest` 15, `PromptSecurityRedactTest` 18, `coreMiddlewareTest` 17 all green.
`spotlessCheck` passes.

I verified the new cache test actually detects the regression rather than just passing: re-keying the
cache per-instance makes it fail with *"two instances ... must share one cache entry"*. Its timing
bound is generous by design — off-AWS the first call spends seconds in timeouts and the second
returns immediately from the negative cache, while on AWS both are fast, so it can't flake either
way.

#249 has been rolled forward onto this head and re-verified.

dougcain:
@lmajano Heads-up before you look again: **the credential-resolution path has changed materially
since `2133508`**, the commit you last reviewed. Head is now `57ba46a`. I'd rather flag that than
have you re-read a diff you'd already signed off.

Two review passes found real P1s in code I had previously reported as fixed — including in the
change your own lock note prompted. Summarised so you can go straight to the parts worth your
judgement.

### Your non-blocking lock note turned out to matter more than either of us thought

It led to #258, and then the fix for #258 introduced two worse problems, both now fixed:

- **The lock became JVM-wide and was held across the metadata I/O.** Keying it on the credential
  source (instead of the old per-instance UUID) meant every Bedrock request in the runtime
  serialized through one exclusive lock, held across ~5s of container+IMDS HTTP plus an untimed
  token-file read. Past the 10s timeout it throws `type="lock"`, which `RetryMiddleware`'s
  `nonRetryableTypes` doesn't cover — so it would sleep and re-queue onto the same congested lock.
  Now: readonly lock for the cache read, resolution **outside** any lock, exclusive lock only for the
  write-back. Two threads racing a cold cache may each pay one round-trip; that's deliberate, and I'd
  welcome your view on the trade.
- **The shared negative cache could turn a cold-start blip into a 60s outage.** One failed
  resolution served empty credentials — a guaranteed 403 — to every request for the full window,
  where per-instance state had let the next request just retry. Now two windows: 60s while usable
  credentials remain (pure refresh suppression), 5s when there's nothing to serve.

### Credential correctness (all found post-review, all fixed)

- The atomic-credential set gated on `keyExists("awsAccessKeyId")`, which **blanked an
  environment-supplied secret** whenever only an access key id was configured — a setup that worked
  before `1ad270e`. It now requires a complete explicit pair; partial overrides apply per key.
- `credSource` switched **wholesale** to the nested `apiKey` struct that `aiService()` injects,
  discarding the caller's own explicit region and key pair. An explicit `eu-west-2` pair could sign
  as the module's `us-east-1` account. The nested struct now fills only what the caller didn't state.
- Explicit per-call options were losing to `settings.providers.Bedrock.options`, because
  `BaseService.configure()` appends them with `override=true` — and it aliases then overwrites the
  caller's struct in place, so the caller's values have to be snapshotted before delegating.
- `loadContainerCredentials()` preferred `_FULL_URI` while the cache key keyed on `RELATIVE_URI`, so
  with both set we fetched from one endpoint and cached under the other. AWS specifies RELATIVE wins.
- `isCredentialCacheValid()` accepted a container endpoint answering `200 {}`: empty keys plus a
  synthesized 60-minute expiry read as valid, suppressing the IMDS fallback and signing with an empty
  secret for an hour. Both loaders now return null on a credential-less 200.
- The cache key no longer includes region (unvalidated caller input could mint unbounded entries each
  holding a live secret), and `resolveAwsCredentials()` returns a copy rather than the live shared
  struct.

### One change outside Bedrock, worth knowing about

`firstHeaderValue()` moved to `BaseService` and is now used for **its own** `retry-after` read. The
array-shaped-header bug you flagged in Bedrock's 429 handlers was still live in
`sendChatRequest()` for every provider routing through it — OpenAI, Grok, Groq, Mistral, DeepSeek,
OpenRouter, Perplexity, MiniMax, HuggingFace, DockerModelRunner, OpenAICompatible, Cohere, Gemini.
Bedrock's duplicate copy is gone; it inherits the shared one. Say the word if you'd rather that
landed separately from this PR.

Also fixed: the stream accumulator now guards on `choices.len()`. I verified in BoxLang that
`first()` on an empty array throws, and the final invocation-metrics event is a usage-only chunk with
`choices: []` — the throw was swallowed at debug level, so the caller's stream callback silently
never received that chunk. `AiAgent` had the identical unguarded read, and fixing Bedrock's alone
just moved the swallowed throw there.

### State

- **`57ba46a`** — BedrockServiceTest 58 (57 passed, 1 skipped: live call, no valid credentials), plus
  ProviderCapabilities 19, ReasoningNormalization 8, RunControl 12, coreMiddleware 17, OutputGuard 9,
  MockService 13. `spotlessCheck` and `git diff --check` pass. The 7 `OpenAITest` failures are live
  calls with a dummy API key.
- #249 is rebased on this head and re-verified (279 tests across the affected suites, 0 failures
  beyond the same dummy-key live call).
- #258 closed. #259 (consolidating `AwsCredentialProvider`'s separate chain — it's live via
  `OpenSearchVectorMemory`) and #257 (per-chunk streaming guard seam) remain open and deliberately
  out of scope here.

One honest observation for your release decision: two consecutive review rounds found P1s in this
area, and the pattern is consistent — credential precedence has several interacting sources of truth
(`arguments.options`, `variables.options`, the nested `apiKey` struct, environment, module settings).
If you'd prefer to release without this and let it land after a dedicated pass on that precedence
chain, that's a reasonable call and I won't argue it.

dougcain:
@lmajano Rebased on `development` and retested as you asked. Head is `c943c99`, MERGEABLE. Your
`null`-literal note was addressed back in `8da4af2`.

### Read the red checks before judging them — both are environmental

- **`Tests bx-ai`** — this PR is from a fork, so the run gets no repo secrets. 153 of the 154 failures
  are missing-API-key (`aiEmbedTest`, the vector-memory integrations, `ClaudeTest`, `GeminiTest`, …),
  and `SummaryMemoryTest`'s two are the same cause — its summarization makes a live LLM call, and it
  passes on `development` where secrets exist. **Zero Bedrock failures.**
- **`Checks Source Code Formatting`** — shows 24h because it never got a runner ("exceeded the maximum
  execution time while awaiting a runner"). Not a spotless violation; `spotlessCheck` passes locally.

### What changed since the `2133508` you reviewed

All of it is in the credential path — that's the area worth your eyes, not the parity work you already
approved.

**Your lock note turned out to matter more than either of us thought.** It led to #258, and the fix for
#258 introduced two worse problems, both now fixed: the lock became JVM-wide and was held across ~5s of
container+IMDS HTTP (and `RetryMiddleware` doesn't treat the resulting `type="lock"` throw as
non-retryable, so it would re-queue onto the same lock); and the now-shared negative cache could serve
empty credentials — a guaranteed 403 — to every request for a full minute after one cold-start miss.
Reads now take a readonly lock, resolution happens outside any lock, and the negative window is short
when there's nothing usable to serve.

**Credential precedence is now strict and atomic.** Sources rank explicit per-call arguments →
`settings.providers.Bedrock.options` → module-global `apiKey` → environment, and whichever source
supplies a complete key pair supplies the session token too, or leaves it empty. Four separate bugs fed
into this: an explicit pair inheriting an unrelated `AWS_SESSION_TOKEN`; a configured access key
blanking an environment secret; provider settings losing to the global `apiKey`; and `credSource`
switching wholesale to the injected `apiKey` struct. All have regression tests.

**Container endpoint hardening.** The request carries a bearer token, so the endpoint is validated now:
plain HTTP only for loopback and the ECS/EKS link-local addresses, HTTPS otherwise. The host is parsed
properly — `http://127.0.0.1:80@evil.example/creds` was read as `127.0.0.1`, and `http://[::1]/` as
`[`. `RELATIVE_URI` must be a path with no userinfo since it's concatenated onto the fixed ECS address,
and it now correctly wins over `FULL_URI` per the AWS spec.

**One change outside Bedrock.** `firstHeaderValue()` moved to `BaseService` and now normalizes its own
`retry-after` — the array-shaped-header bug you flagged was still live there for all thirteen providers
routing through `sendChatRequest()`. Happy to split that into its own PR if you'd rather it didn't ride
along.

### Tests

`BedrockServiceTest` 61 — 60 passed, 1 skipped (live AWS call, no valid credentials; it now reports
*skipped* rather than falsely passing). `spotlessCheck` and `git diff --check` pass.

### Related

#249 is stacked on this head and green (138 tests). #260 fixes the *other* credential chain —
`AwsCredentialProvider`, which `OpenSearchVectorMemory` uses: its `parseCredentialsFile()` called
`chr()`, which isn't a BoxLang function, so AWS profile-file credentials have never worked, and it sent
no `Authorization` header so EKS Pod Identity couldn't work either. That one is based on `development`
and independent of this stack. #257 and #259 track the remaining follow-ups.

### One honest note for your release decision

Four review rounds have now found P1s in the credential-precedence code, each in the previous round's
fix. The cause is consistent: that path has several interacting sources of truth
(`arguments.options`, `variables.options`, the nested `apiKey` struct, provider settings, environment).
If you'd rather release without this and let it land after a dedicated pass on that chain, that's a
reasonable call and I won't argue it.

Also: one small thing worth aligning at some point — `development`'s Claude sync fix selects the
**first** text block while Bedrock joins **all** of them. Mine is arguably more correct, but the two
providers now differ.

PR #235, opened 2026-08-03
🔐 Fix Gemini API key leaking into logs in plaintext

lmajano:
Gemini authenticates with a `?key=` QUERY PARAMETER. The per-request key was
written into variables.chatURL - instance state on a long-lived shared service -
which BaseService then printed verbatim as `Endpoint: ...` whenever logRequest
or logRequestToConsole was enabled. A live API key therefore landed in the `ai`
log file and on stdout during ordinary debugging.

Header-authenticated providers were never affected: buildRequestLog only ever
records header NAMES, never values. Bedrock already did this correctly, logging
SET/MISSING instead of credentials.

Fix, in two independent layers:

1. The key-bearing URL is now request-local and passed straight to the
   transport; only the key-free endpoint is stored on the service instance.
   This also removes a cross-request hazard - a shared service instance could
   previously expose one caller's key to another under concurrency.

2. Defense in depth: new PromptSecurity::redactURLSecrets() masks
   key/api_key/token/access_token/secret/password/sig/signature query-param
   values in any logged endpoint, covering every provider including any future
   key-in-URL one. Non-secret params (alt=sse) are preserved so logs stay
   useful. Placed on PromptSecurity so it is publicly unit-testable and
   reusable rather than a private provider helper.

Verified end-to-end with logging enabled: the key appeared TWICE in output
before the fix and ZERO times after. The source guards in ApiKeyUrlLeakTest
were each confirmed to fail when the bug is reintroduced.

Note: GeminiTest's 5 failures in a keyless environment are pre-existing and
identical on clean development - they are live tests. That they still reach
Google and return "API key not valid" (rather than a missing-key error)
confirms the key is still transmitted correctly.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Ks4wJDJaJiMCkmcFnEcipQ


lmajano:
@copilot please fix the merge conflicts in this pull request.

lmajano:
@copilot please fix the merge conflicts in this pull request.

PR #253, opened 2026-08-16
Normalize reasoning onto the shared provider envelope

lmajano:
## Description

Reasoning-capable models could always be **enabled** — `params` passes through to the provider body verbatim, so this already reached the API today:

```javascript
params: { thinking: { type: "enabled", budget_tokens: 10000 } }   // Claude
params: { reasoning_effort: "high" }                              // OpenAI
```

But what came back was **parsed out and silently dropped**. `ClaudeService`'s stream reader only handled `text_delta` / `input_json_delta` / `tool_use`, and `BaseService` only extracted `delta.content` / `delta.tool_calls`. A `thinking_delta` chunk matched no branch and was discarded — so you paid for thinking tokens and never saw them.

Reasoning now surfaces on the **same OpenAI envelope every provider already normalizes onto**:

| | Key |
|---|---|
| Streaming | `choices[].delta.reasoning` |
| Synchronous | `choices[].message.reasoning` |

Consumers read it identically regardless of backend — no provider branching:

```javascript
var reasoning = chunk.choices?.first()?.delta?.reasoning ?: ""
```

## Every provider is covered

All 21 provider classes were audited for how each obtains its response. There are exactly three routes, and all three are handled:

| Route | Providers | How |
|---|---|---|
| Delegate to `super.chat()`/`super.chatStream()` | Grok, Groq, Mistral, DeepSeek, OpenRouter, Perplexity, MiniMax, HuggingFace, DockerModelRunner, OpenAICompatible | Inherit `BaseService` normalization — verified, no change needed |
| Call the inherited transport directly | Cohere, Gemini | Same |
| Build their own chunks/responses | Claude, Ollama, Bedrock, Mock | Explicit mapping added |

Native spellings mapped onto the one key: Anthropic `thinking_delta`, DeepSeek `reasoning_content`, Ollama `message.thinking`, Claude-on-Bedrock `delta.thinking`, plus OpenAI-shaped Bedrock models in either spelling.

## Two real bugs found while doing this

Both meant enabling reasoning **actively broke the provider**, not merely hid the thinking:

1. **Claude sync returned an empty answer.** `chat()` read the answer as `result.content.first().text`. With extended thinking on, Anthropic puts `thinking` blocks *before* the text block — so that read was null and **every** sync return format (`single`, `json`, `xml`, `structuredOutput`) silently produced nothing. Now selects the first *text* block.
2. **Bedrock discarded thinking-only chunks.** `transformStreamChunk()` bailed on any chunk with neither content nor a finish reason — precisely what every chunk looks like while a model is still reasoning. Reasoning is now part of that guard.

## Design decisions worth reviewing

- **Absence is normal, never an error.** A provider or model with no reasoning simply omits the key; `?: ""` degrades silently.
- **Deliberately NOT a capability interface.** Reasoning varies per **model** (Sonnet vs. Haiku, gpt-5 vs. gpt-4o), and the existing `capabilities/` mechanism is per-*provider* marker interfaces — an `IAiReasoningService` would be structurally unable to express that and would misreport half the time.
- **Reasoning never enters memory.** Kept strictly separate from `content` and never folded into the persisted assistant message — replaying a model's private thinking back to it as if it had said it changes its behavior next turn. `AiAgent.stream()` was already correct by construction (it only accumulates `delta.content`); the invariant is now explicit and locked in by a test.
- **`MockService` applies the identical normalization in its own overrides.** It replaces the transport wholesale, so without that, mock and production would follow different contracts and every test written against the mock would assert something production doesn't do.
- **Anthropic's `signature_delta` and redacted thinking blocks are not surfaced** — no cross-provider equivalent.

### Known asymmetry, left deliberately

Claude's sync `returnFormat: "raw"` returns Anthropic's **native** envelope, not an OpenAI `choices[]` shape. That's pre-existing behavior and converting it would be breaking, so reasoning is attached at `result.reasoning` there while streaming uses the normalized `delta.reasoning` like everywhere else. Flagging it rather than silently papering over it — happy to unify in a follow-up if that's wanted.

## MockService

Reasoning can be scripted, emitted **before** content just like a real provider, so reasoning-aware consumers are testable offline with no reasoning-capable model and no API key:

```javascript
providerOptions: { responses: [
    { content: "Rayleigh scattering.", reasoning: "Consider light wavelengths." }
] }
```

## Type of change

- [x] Bug Fix (the two above)
- [x] New Feature
- [x] This change requires a documentation update (done — readme + changelog)

## Checklist

- [x] My code follows the style guidelines of this project (`spotlessCheck` passes locally)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective and that my feature works
- [x] New and existing unit tests pass locally with my changes

## Test plan

`ReasoningNormalizationTest` — 7 tests, all MockService-backed (no network, no API keys):

- [x] synchronous reasoning lands on `message.reasoning`, separate from content
- [x] a turn with no reasoning omits the key — absence is normal, not an error
- [x] a provider-native `reasoning_content` is normalized on the sync path (uses a **raw** completion that bypasses the mock's own message builder, so it exercises the normalization rather than the mock writing the key directly — verified this genuinely fails without the fix)
- [x] Claude picks the first **text** block, so thinking blocks don't blank the answer (asserts both the corrected read *and* that the old read returned nothing, so the bug can't quietly return)
- [x] streaming emits reasoning on `delta.reasoning`, ordered **before** any content delta
- [x] streamed reasoning reaches the caller but is **never** persisted to agent memory (asserts the answer *is* stored while the thinking is *not*)
- [x] every streamed chunk conforms to the shared envelope, reasoning included

Regression sweep — `providers.*` + `runnables.*` + `middleware.*`: **341 tests / 63 failures**, against **340 / 63** for the identical selection on this branch without these changes. One added test, zero new failures. All 63 are pre-existing live-provider tests failing on `Incorrect API key provided: your-api-key` in this sandbox — CI with real keys is the check that matters here, which is what this PR is up for.


lmajano:
Pushed a follow-up commit (`d87cd22`) fixing a real gap in the original — worth calling out since the first commit's docs overclaimed.

**What was wrong:** streaming was genuinely covered (OpenAI's `chatStream()` routes through `BaseService.sendStreamRequest()`, so `normalizeReasoningDelta()` reached OpenAI and every provider extending it). But the **synchronous `chat()` path had no normalization at all** — it returned the raw provider response, so a real provider answering with the native `reasoning_content` spelling would *not* surface as `message.reasoning`, despite the readme and changelog claiming it would.

The original sync test passed only because `MockService` writes `message.reasoning` directly — it never exercised the normalization itself.

**Fix:**
- Added `normalizeReasoningMessage()`, called from `BaseService.sendChatRequest()` before the `onAIChatResponse` announce, so listeners and callers both see the standard shape.
- `MockService` replaces the transport wholesale, so it now applies the identical normalization in **both** of its own overrides. Without that, mock and real providers would follow different contracts, and every test written against the mock would assert something production doesn't do.

**New test:** uses a RAW scripted completion carrying `reasoning_content` (bypassing the mock's own message builder), which is the shape a real provider actually returns. I verified it genuinely **fails without the fix and passes with it**, rather than assuming — the earlier overclaim came from not doing that check the first time.

Suite is now 6 tests, all passing. Regression sweep re-run: 231 tests / 12 failures, identical to the baseline on clean `development` (all 12 are pre-existing live-provider tests failing on missing API keys in this sandbox).

---

lmajano:
All providers now covered (`8560e37`). I audited all 21 provider classes rather than guessing — there are exactly three routes a provider takes to get its response, and all three are handled:

| Route | Providers | Status |
|---|---|---|
| Delegate to `super.chat()`/`super.chatStream()` | Grok, Groq, Mistral, DeepSeek, OpenRouter, Perplexity, MiniMax, HuggingFace, DockerModelRunner, OpenAICompatible | Inherit `BaseService` normalization — verified, no change needed |
| Call the inherited transport directly | Cohere, Gemini | Same |
| Build their own chunks/responses | Claude, Ollama, Bedrock, Mock | Explicit mapping added |

### Two real bugs found doing this

Both meant enabling reasoning **actively broke the provider**, not merely hid the thinking:

**1. Claude sync was blanked by thinking blocks.** `chat()` read the answer as `result.content.first().text`. With extended thinking on, Anthropic returns `thinking` blocks *before* the text block — so that read was null and **every** sync return format (`single`, `json`, `xml`, `structuredOutput`) silently returned an empty answer. Now selects the first *text* block. The test asserts both that the corrected read returns the answer and that the old read returned nothing, so it can't quietly regress.

**2. Bedrock discarded thinking-only chunks.** `transformStreamChunk()` bailed on any chunk with neither content nor a finish reason — which is exactly what every chunk looks like while the model is still reasoning. Reasoning is now part of that guard.

Also mapped: Ollama's `message.thinking` (deepseek-r1 and friends), Claude-on-Bedrock's `delta.thinking`, and OpenAI-shaped Bedrock models in either native spelling.

### Verification

7 reasoning tests, all passing. Regression sweep across `providers.*` + `runnables.*` + `middleware.*`: **341 tests / 63 failures**, against **340 / 63** for the identical selection on this branch without these changes — one added test, zero new failures. All 63 are pre-existing live-provider tests failing on missing API keys in this sandbox.

---

lmajano:
## CI status

Failures are down from **12 on the base commit (`3c3433b4`) to 2–4** here, depending on the run — two runs fired on `288fb811` and reported 2 and 4 respectively, which is itself the tell that what remains is live-provider flake.

Two real bugs CI caught along the way, both fixed:

**`7fb92ef` — Ollama streaming was broken by this PR's first commit.** Every chunk logged:

```
Error emitting normalized chunk: The key [reasoning] was not found in the struct.
Valid keys are ([usage, content, toolCalls, finishReason])
```

`BaseService.updateStreamState()` assumed the accumulator struct always carries a `reasoning` key, but providers that override the stream loop build their own — Ollama's declared four keys. Worse than a noisy log: the throw was swallowed by the emit `try/catch`, which then skipped the rest of `updateStreamState()` for that chunk, silently stopping *content* accumulation for the whole stream. Fixed by seeding the key instead of assuming it, plus declaring it in Ollama's own struct. Regression test verified to fail against the previous code.

**`288fb81` — the three OpenAI tool tests were hitting an endpoint constraint, not a bug in this PR.** OpenAI rejects function tools alongside active reasoning on `/v1/chat/completions`:

> Function tools with reasoning_effort are not supported for gpt-5.6-luna in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.

The provider default model is a reasoning model, so this fires even though nothing in the module sends `reasoning_effort` — the model reasons by default. These tests exist to exercise tool calling, so they now pin `gpt-4o`. All three are green.

Worth flagging separately from this PR: **this affects users, not just tests.** Anyone passing `tools` on OpenAI's default model hits it. The readme's reasoning section now documents both caller workarounds (name a non-reasoning model, or set `reasoning_effort: "none"`). The real fix is Responses API support, which is deliberately out of scope here.

Remaining failures are environmental and unrelated to this diff:

- `HuggingFaceTest > Test HuggingFace AI chat` — account inference quota exhausted
- `PerplexityTest > Test Perplexity AI` — HTTP 429 rate limit
- `OllamaTest > Test XML response` — `qwen3:0.6b` returning unparseable XML; this is the **synchronous** path, and every line of this PR's Ollama change is inside `sendStreamRequest`

---

Issue #198, opened 2026-06-18
Enable structured data response with Amazon Bedrock


dandyfopper:
## Summary

Add structured-data-response support for Anthropic Claude models on **AWS Bedrock** and the **native Anthropic Messages API**. Today `aiChat(...).schema(...)` / structured output works for OpenAI, Ollama, and Cohere (they inject the schema via `response_format`/`format`), but for Claude it is silently unsupported: the schema is never sent to the model, so Claude returns prose and `populateStructuredOutput()` fails when it tries to parse it as JSON. This makes structured output unusable against Bedrock Claude (and native Claude), forcing callers to fall back to a local model.

## Detailed Description

**Problem**

`BedrockService.transformRequestForClaude()` (and `ClaudeService.chat()`) build the request from messages/params/system/tools only — they never consult `chatRequest.isStructuredOutput()`. Neither the Bedrock Claude Messages API nor the native Anthropic API supports OpenAI-style `response_format`, so there is no schema-enforcement path. Result:

- Schema is never transmitted → Claude answers in natural language.
- `populateStructuredOutput()` then runs `jsonDeserialize()` on prose → opaque "Unrecognized token 'Here'" parse error.
- Works on a local Gemma model (Ollama/DockerModelRunner inject the schema) but not Bedrock/Claude — confirmed in the `sentiment.cfm` integration.

**Desired behavior**

`structuredOutput()` / `schema()` should return a populated struct/class/array from Claude on Bedrock and from the native Anthropic API, with parity to the OpenAI path, and should fail loudly (not silently) when the model can't comply.

**Reference**: AWS — "Structured data response with Amazon Bedrock: prompt engineering and tool use"
https://aws.amazon.com/blogs/machine-learning/structured-data-response-with-amazon-bedrock-prompt-engineering-and-tool-use/

## Possible Implementation Ideas

Use the **forced tool-use** pattern (Anthropic's recommended approach), applied in both `BedrockService` and `ClaudeService`:

1. When `isStructuredOutput()` is true for a Claude model, inject a synthetic tool (`structured_output`) whose `input_schema` is the requested JSON schema (`SchemaBuilder::cleanSchema(...)`), and pin `tool_choice` to it. Real user tools remain in the packet but won't be called while `tool_choice` is forced.
[...]

PR #271, opened 2026-09-06
Bedrock/Cohere streaming tool calls + batch HITL resume, FlightRecorder pluggable storage, streaming structured output (#248, #170, #198)

### LINKED ISSUE #170: FlightRecorderMiddleware: Support custom storage backends via callback / observe mode
## Summary

Allow `FlightRecorderMiddleware` to persist tapes to custom storage backends (database, cloud, message queue) in addition to — or instead of — local JSON files. Today the middleware is hardwired to file-based storage via a private `_saveSnapshot()` method, forcing applications that need DB-backed auditing to duplicate the entire capture pipeline in a separate middleware.

## Detailed Description

`FlightRecorderMiddleware` has a clean, well-tested capture pipeline: `wrapLLMCall` and `wrapToolCall` intercept every interaction, append it to an in-memory tape, and flush to disk via `_saveSnapshot()`. The tape format, sequencing, crash-safe incremental writes, and replay logic are all excellent.

The problem is that `_saveSnapshot()` is private and writes exclusively to a JSON file. Applications that need to store tapes in a database — for multi-tenant isolation, retention policies, full-text search, or audit compliance — cannot hook into this pipeline. They must build a completely separate middleware from `BaseAiMiddleware` that reimplements the same `wrapLLMCall`/`wrapToolCall` capture pattern, duplicating ~80% of FlightRecorderMiddleware's logic.

This is what we had to do: we wrote an `AuditRecorderMiddleware` (~120 lines) that extends `BaseAiMiddleware` and reimplements the same interaction capture from `wrapLLMCall`, but saves to MariaDB via a service layer instead of to disk. Almost all of it is duplicated from FlightRecorderMiddleware.

Additionally, there is no "observe" mode — `passthrough` doesn't accumulate the tape, `record` always writes files, and `replay` reads from files. There's no way to accumulate the tape in memory without file I/O, which would be useful for DB-only storage or streaming to an external system.

### Idea 1: `storageCallback` parameter

Add a callback that fires after each interaction is appended to the tape. This is the least-invasive change — one new constructor parameter, a few lines in `_appendInteraction()`:

```boxlang
// New constructor parameter
function init(
    // ... existing params ...
    function storageCallback = ""
)

// In _appendInteraction(), after the existing _saveSnapshot() call:
if ( isCustomFunction( variables.storageCallback ) || isClosure( variables.storageCallback ) ) {
    try {
        variables.storageCallback( variables._tape.copy(), arguments.interaction )
    } catch ( any e ) {
        writeLog( text: "FlightRecorder storageCallback error: #e.message#", type: "warning", log: "ai" )
    }
}
```

Usage:

```boxlang
model.withMiddleware(
    new FlightRecorderMiddleware(
        mode            : "record",
        storageCallback : ( tape, interaction ) => {
            myDbService.saveTape( tape: tape, label: "auto" )
        }
    )
)
```

### Idea 2: Add an "observe" mode

A fourth mode alongside passthrough/record/replay that accumulates the in-memory tape (like `record`) but does NOT write files. Combined with `storageCallback`, this covers DB-only use cases without any file I/O:

```boxlang
new FlightRecorderMiddleware(
    mode            : "observe",
    storageCallback : ( tape, interaction ) => {
        auditService.saveToDatabase( tape )
    }
)
```

The implementation would be minimal — `wrapLLMCall` and `wrapToolCall` already check `variables.mode` with an if/else chain; `observe` would follow the same path as `record` but skip the `_saveSnapshot()` call.

### Idea 3: Make `_saveSnapshot()` overridable

Change `_saveSnapshot()` visibility from `private` to allow subclass overrides:

```boxlang
class extends="FlightRecorderMiddleware" {
    function _saveSnapshot() {
        dbService.saveTape( tape: variables._tape )
    }
}
```

This is the most OOP approach and requires the smallest code change, but is less flexible than the callback (one storage mechanism per subclass, no composition).

### Recommendation

Ideas 1 + 2 together would cover the full range of use cases with no breaking changes. The callback handles custom storage, and `observe` mode eliminates unnecessary file I/O for applications that only need DB persistence. Idea 3 is a good complementary change if the team prefers inheritance-based extensibility.

### LINKED ISSUE #248: Bedrock and Cohere streaming have no tool-call support, so streaming HITL/batch approvals silently no-op
## Summary

`BedrockService.chatStream()` and `CohereService.chatStream()` have no tool-call handling
at all. Tool-use blocks are parsed out of the event stream and discarded, `beforeToolCall`
never fires, and no follow-up turn is issued. Because 3.4.0's batched tool-call approvals
build on that tool loop, streaming HITL is a silent no-op on both providers — and
`AiAgent.resumeStream()` will replay the LLM without executing the approved tools.

This is **pre-existing**, not a 3.4.0 regression. The 3.4.0 changelog is accurate that
streaming batching "covers OpenAI and Claude (the only two with streaming tool-call
support today)". This issue is to close that gap, and to make the current behaviour loud
rather than silent in the meantime.

## Current state

| Provider | sync `chat()` batch approvals | streaming `chatStream()` batch approvals |
|---|---|---|
| OpenAI | ✅ `OpenAIService.bx:118-120`, `370-416`, `resumeToolBatch()` `:470` | ✅ `:582-584`, `:776-818`, `resumeToolBatchStream()` `:873` |
| Claude | ✅ `ClaudeService.bx:596-598`, `815-845`, `:1105` | ✅ `:105-107`, `:428-469`, `:993` |
| **Bedrock** | ✅ complete and faithful | ❌ **no tool handling of any kind** |
| **Cohere** | ✅ `CohereService.bx:157-159`, `372-402`, `:531` | ❌ no tool loop (`chatStream()` `:647`, `toolCalls: []` hardcoded `:897`) |

## Reproduction / evidence

A Bedrock **streaming** call whose model requests 2+ tool calls needing HITL approval does
not suspend, does not error, and does not run the tools. It silently returns only the text
deltas preceding the tool call.

Trace (line numbers on `upstream/development` @ `13ef57c`):

1. `chatStream()` passes the capability gate (family `claude` is allowed) and sends `tools`.
2. The event stream reaches `transformStreamChunk()`. Tool-use arrives as
   `content_block_start`, which hits:
   ```
   default:
       // Skip other event types like message_start, content_block_start
       return javacast( "null", "" );
   ```
3. `beforeToolCall` is never fired anywhere in the streaming path — the only `tool` token
   in that whole range is the hardcoded `"toolCalls": []` handed to `afterLLMCall`.
4. No suspension, no recursion for a follow-up turn. The caller's callback sees text only.

## Related latent bug — `resumeStream()` mis-resumes

`AiAgent.bx:1114-1120` passes `_resumeContext` into `stream()` regardless of provider, and
Bedrock's `chatStream()` never reads it. This is reachable **today**:

1. A sync Bedrock `chat()` suspends a tool batch for approval.
2. The application calls `agent.resumeStream()` instead of `agent.resume()`.
3. The LLM is replayed from scratch, the approved tools never execute, and no error is
   raised — the user believes the approved action was taken.

## Proposed work

- Parse `content_block_start` / `content_block_delta` tool-use blocks in
  `BedrockService.transformStreamChunk()` and accumulate them into normalized tool calls.
- Wire the streaming tool loop through the middleware pipeline (`beforeToolCall` /
  `afterToolCall` / `wrapToolCall`) as `ClaudeService.chatStream()` does.
- Add `afterToolBatch` / `defer()` participation and `resumeToolBatchStream()`, mirroring
  `ClaudeService.bx:993`.
- Same for `CohereService.chatStream()`.
- Fix `resumeStream()` so it can no longer silently replay against a provider whose
  streaming path lacks tool support.

**Note on sequencing:** the `resumeStream()` mis-resume above is reachable today and fails
silently, whereas the streaming tool loop is days of work across two providers. If the
streaming work is going to sit, that one is worth pulling out and fixing on its own — happy
to send it as a standalone PR if useful.

**Acceptance:**
- `SuspendResumeIntegrationTest` gains streaming suspend/resume/reject/cancel coverage for
  Bedrock and Cohere, matching the existing OpenAI/Claude cases.
- A streaming call with 2+ approval-requiring tool calls suspends **once** as a batch.
- `resumeStream()` on a batch checkpoint completes without replaying the LLM.

## Related: Cohere-on-Bedrock is not functional end to end

Separate from streaming, plain Cohere-on-Bedrock chat does not work at all.
`transformRequestForModel()` deliberately routes family `cohere` to
`transformRequestForClaude`, so Command R receives `{ anthropic_version, messages,
max_tokens }` where its InvokeModel API expects `{ message, chat_history, max_tokens }`.
AWS rejects the body, so the call 400s before any response handling runs.

The response half has been fixed (a dedicated `transformResponseFromCohere` covering both
the Command R `{ text }` and legacy Command `{ generations[] }` shapes), but it is only
reachable once a matching request transform exists. Needs `transformRequestForCohere()`
plus a Cohere branch in `transformStreamChunk()`; the three switches
(request / response / stream) should agree on the family again afterwards.

## Out of scope

Bedrock provider parity (#227 / PR #233) is tracked separately, as is the Bedrock
response-path work (the Cohere response transform, and teaching
`PromptSecurity::getResponseText`/`setResponseText` Bedrock's native Titan/Llama/Mistral
shapes so `OutputGuardMiddleware` stops silently no-op'ing on those families). This issue
is only about streaming tool-call capability.

dougcain:
## Summary

Closes #170, closes #248, and completes the follow-ups noted on #198. Also fixes an unfiled `ClosureTool` bug found while writing the streaming tests.

### #170 — FlightRecorderMiddleware pluggable storage
- `storageCallback( tape, interaction )` init param, fired after each interaction in `record` and `observe` modes
- new `observe` mode: in-memory tape, no file I/O
- `_saveSnapshot()` is public so subclasses can override persistence
- unknown mode throws `FlightRecorder.InvalidMode` at init instead of falling into replay
- `wrapLLMCall` honours `ctx.stream` (see below) — raw stream bodies are taped/replayed, emit-based seams pass through

### #198 follow-ups — structured output
- `ClaudeService.chatStream()` and `BedrockService.chatStream()` now inject the forced `structured_output` tool, accumulate it separately from real tool calls, and return the populated value; `StructuredOutputError` if the block never arrives
- `populateStructuredOutput()` extracts JSON via `JSONExtractorTransformer`, so prose-wrapped replies populate on OpenAI/Gemini/Cohere/Ollama; `NoJSONFound` → `StructuredOutputError`

### #248 — Bedrock / Cohere streaming tool calls + batch HITL
**Bedrock**
- streaming tool loop: `tool_use` accumulation, `beforeToolCall`/`wrapToolCall`/`afterToolCall`, `afterToolBatch`/`defer()`, `resumeToolBatchStream()`, `_resumeContext` honoured; follow-up turns stream
- event-stream header decoding: exception frames (`throttlingException`, `modelStreamErrorException`, …) now throw `BedrockStreamError` instead of silently ending the stream; truncated/malformed frames fail loud
- OpenAI family (`gpt-oss` and the catch-all): full chat-completions body (`tools`, `tool_choice`, `response_format`, `reasoning_effort`, `stop`, `max_completion_tokens`); tools in chat and stream; native `response_format` structured output; `"openai"` removed from `NO_TOOL_SUPPORT_FAMILIES` (legacy `cohere.command-text` still gated by id)
- `transformRequestForCohere()` (Command R + legacy) and a Cohere stream branch — request/response/stream switches agree again
- `cache_control` preserved on system prompt (array of blocks when present)
- truncated tool-argument buffers throw instead of running tools with `{}`
- stream HTTP transport wrapped in `wrapLLMCall` with `ctx.stream = true`, `ctx.transport = "bedrock-event-stream"`; non-string returns from middleware throw a clear error

**Cohere**
- `chatStream()` handles v1 `tool-calls-chunk` / `tool-calls-generation` / `stream-end`, runs the tool loop with `afterToolBatch`/`defer()`, adds `resumeToolBatchStream()`; SSE transport wrapped in `wrapLLMCall`

**AiAgent**
- `resumeStream()` throws `AiAgent.StreamingResumeUnsupported` for providers whose stream path can't honour a batch resume, checked before `clearState()` and only on the batch path. Detection via `supportsStreamingToolResume()`; `GeminiService` reports false

### ClosureTool
- `doInvoke()` mutated the caller's live `tool_use.input` (coercion + `_chatRequest` injection), leaking `_chatRequest` into the next request body and creating cycles that overflowed the stack on Bedrock streaming follow-ups. Now works on a `structCopy`.

## Tests
- `SuspendResumeIntegrationTest` +10: Bedrock/Cohere streaming suspend-once-as-batch, agent-level `resumeStream()` approve/reject/edit/cancel, no LLM replay, `_chatRequest` leak guard
- `BedrockServiceTest` 70 → 93 with real binary event-stream frame fixtures (headers, exception frames, tool accumulation, OpenAI/Cohere shapes, truncation)
- new: `CohereStreamingToolsTest`, `AiAgentResumeStreamGuardTest`, `StructuredOutputParsingTest`; `FlightRecorderMiddlewareTest` 14 → 22; `ClosureToolTest` +2
- full suite: 0 logic failures (remaining failures are missing API keys / local services)

## Not verified live
No AWS credentials in the build environment. gpt-oss `tools`/`response_format` support is inferred from AWS's Bedrock OpenAI page deferring to the chat-completions spec; Bedrock Cohere stream event names come from Cohere's v1 docs. Would appreciate a run against real `eu.anthropic.*`, `openai.gpt-oss-*` and `cohere.command-r*` before release.

## Known leftovers
- sync `BedrockService.resumeToolBatch()` lacks the max-interactions check the stream variant now has (pre-existing)
- no Converse/ConverseStream, CountTokens, or in-provider throttle retry yet


dougcain:
Pushed fixes for three review findings (offline-reproduced):

1. **Cohere streaming ran tools with `{}` on truncated arguments** — `finalizeToolCallBuffers()` now delegates to `parseStreamToolArguments()`, which throws `CohereStreamError` ("tool was NOT executed") before the batch runs, matching the Bedrock behaviour.
2. **Cohere follow-up turn restarted from the original request** — `sendToolResultsFollowUp()` is now a loop: a follow-up that requests another batch is executed in place with `tool_results` accumulated across rounds and the `MaxInteractionsExceeded` guard applied; a later-round suspension carries `priorToolResults` so resume replays the whole exchange.
3. **FlightRecorder dropped binary Bedrock stream bodies** — `isBinary()` bodies are taped base64-encoded with an `encoding` marker and decoded to bytes on replay.

Tests: +2 `CohereStreamingToolsTest` (truncated args throw / two-round follow-up: A once, B once, 3 LLM calls, round-1 results present in requests 2 and 3), +1 `FlightRecorderMiddlewareTest` (binary round-trip byte-equal, no live call).

dougcain:
Note: a max-effort review of this branch's code found several more issues (Cohere `resumeToolBatch` still restarting via `chat()`, `.filter().first()` throwing on empty tool lists, `AiAgent.resume*` clearing the checkpoint before `resolvePendingActions`, FlightRecorder sync replay returning stream bodies / positional matching, Claude pass-2 ignoring `beforeToolCall` arg replacement). The fixes landed in the stacked PR #273 (`feature/bedrock-converse-api`) alongside the Bedrock ones, since that branch contains this one. If #271 is to merge independently first, say so and I will cherry-pick the non-Bedrock fixes onto this branch.

dougcain:
Cherry-picked the non-Bedrock review fixes onto this branch as its own commit so #271 can merge independently: `AiAgent` resume ordering, Cohere `resumeToolBatch`/`getTool()`/arg-replacement/`middleware_stop`, Claude streaming structured output + pass-2 args, FlightRecorder replay discrimination + strict behaviour + snapshot, `extractJSONContent` catch-scope fix, `ClosureTool` injection gating. Bedrock/Converse fixes stay in #273. Affected suites green (241 pass; only the 7 live-key Claude cases fail), spotless clean. Changelog gains a Migration note: `strict: true` FlightRecorder replays now throw `StreamReplayUnsupported` for untaped stream calls.

lmajano:
@dougcain I love this but can you do smaller PRs so we can trace them easily. Please break them up

dougcain:
Superseded by a split into smaller PRs, as requested: #274 (BoxLang 1.17 pin / 1.14 floor), #275 (#170 FlightRecorder), #276 (#198 streaming structured output), #277 (#248 streaming tool loop + batch resume; independent), #278 (#272 Converse API; stacked on #277). Everything in this PR is in one of those, with review fixes applied. Closing.

PR #273, opened 2026-09-06
Bedrock: Converse / ConverseStream API with InvokeModel fallback (#272)

dougcain:
## Summary

Closes #272. **Stacked on #271** — this branch contains all of #271's commits, so the diff shows both until #271 merges. Merge #271 first. Converse work is `c2d8de4`; later commits are code-review follow-ups (`6960fe2`, `acc3edf`), the #271 back-merge (`542ee31`), and the runtime bump (`5498b46`, `97d7d78`).

`BedrockService` now speaks Bedrock's model-agnostic **Converse / ConverseStream** API by default, with **InvokeModel / InvokeModelWithResponseStream** as a selectable and automatically-applied fallback.

### API selection
- `providerOptions.bedrockApi`: `converse` (default) | `invoke`. Module setting `bedrockApi`, env `BOXLANG_MODULES_BXAI_BEDROCK_API`; precedence per-request > env > application scope > module.
- Automatic fallback to InvokeModel when: model id is on `NO_CONVERSE_MODEL_PATTERNS` (text-completion-only models: `cohere.command-text-*`, `command-light-text-*`, `ai21.j2-*`), `providerOptions.rawBody` is set, or Converse returns a `ValidationException` naming the model/operation as unsupported. Retried once, logged once per model id.
- Fallback is refused when it would be unsafe: after the tool loop has started (messages already carry Converse blocks), after a stream has delivered any chunk to the caller, or for feature-level 400s (`toolChoice`, `toolResult`, images, documents…) where AWS's message is surfaced verbatim instead. Throughput / inference-profile / authorization / throttling messages never trigger fallback.
- The fallback decision is held in a per-request `AiChatRequest.runtimeOptions` slot, never written back into caller-owned `providerOptions`.

### Request (`transformRequestForConverse`)
Content blocks `text` / `image` / `document` / `video` / `toolUse` / `toolResult` / `reasoningContent` (signature preserved) / `cachePoint`; `system[]`; `inferenceConfig`; `toolConfig` + `toolChoice` (`auto`/`any`/`tool`); unknown params (`thinking`, `top_k`, `reasoning_effort`, …) → `additionalModelRequestFields`; `guardrailConfig` / `performanceConfig` as first-class fields (headers kept for the invoke path); `streamProcessingMode` only on `converse-stream`. Consecutive same-role turns are merged (Converse requires alternation); the transform is idempotent on already-Converse-shaped messages so follow-up tool turns pass through.

### Response / stream
[...]

Issue #272, opened 2026-09-06
Bedrock: adopt Converse / ConverseStream API with InvokeModel fallback


dougcain:
## Summary

Move `BedrockService` onto Bedrock's model-agnostic **Converse** / **ConverseStream** runtime API, keeping **InvokeModel** / **InvokeModelWithResponseStream** as a fallback for models or features Converse cannot express.

## Why

`BedrockService` currently hand-builds each vendor's native body (Anthropic Messages for Claude, chat-completions for gpt-oss, `message`/`chat_history` for Cohere, `inputText` for Titan, `prompt` for Llama/Mistral) and parses each vendor's native response and stream. That is six per-family switch arms × {request, response, stream}, and every family gap found so far (#227, #233, #248, #249) has been one of those arms disagreeing with the others. After #271 the tool loop is family-factored, but the transforms are not.

Converse gives one request/response/stream schema for every model: `messages[]` with typed content blocks (`text`, `image`, `document`, `toolUse`, `toolResult`, `reasoningContent`), `system[]`, `inferenceConfig`, `toolConfig` (`tools` + `toolChoice`), `guardrailConfig`, `performanceConfig`, `additionalModelRequestFields` for vendor extras (e.g. Claude `thinking`, `top_k`), and a unified `output.message` / `stopReason` / `usage` / `metrics` / `trace` response. ConverseStream emits unified `messageStart` / `contentBlockStart` / `contentBlockDelta` / `contentBlockStop` / `messageStop` / `metadata` events.

## Scope

**API selection**
- `providerOptions.bedrockApi`: `converse` (default) | `invoke`. Module setting + `BOXLANG_MODULES_BXAI_BEDROCK_API` env var follow the usual priority (env > application > module).
- Automatic fallback to InvokeModel when: the model is on a known no-Converse list, Converse returns a `ValidationException` indicating the model/operation is unsupported, or the request carries an invoke-only escape hatch (`providerOptions.rawBody`). Log the fallback once per model id at `info` on the `ai` log.

**Request transform (`transformRequestForConverse`)**
- messages → Converse content blocks incl. multimodal (`image`, `document` with bytes/base64), assistant `toolUse`, user `toolResult` (status success/error), `reasoningContent` round-trip for Claude thinking.
- system → `system[]` text blocks; `cache_control` → `cachePoint` blocks.
- `inferenceConfig` ← `max_tokens`/`temperature`/`top_p`/`stop`/`stop_sequences`.
[...]

PR #274, opened 2026-09-07
build: pin BoxLang 1.17.0, require >= 1.14.0

dougcain:
## Summary
Build-only. Pins the test runtime to BoxLang **1.17.0** (`gradle.properties`, was 1.13.0) and raises `box.json` `minimumVersion` from 1.8.0 to **1.14.0**.

**Why:** BoxLang ≤ 1.13 resolves an unqualified identifier inside a `catch` body against the class `variables` scope *before* function locals and `arguments` (instance vars, properties and method names all shadow). Fixed upstream in 1.14.0; verified with a 7-case repro on 1.13.0 (broken) and 1.14.0 / 1.17.0 / 1.17.3 (clean). 1.17.0 is the newest 1.17.x with both a `-snapshot` and a release jar published, which the build's branch-based suffix logic needs.

**Only compile break:** `ModuleRecord.moduleConfig` is now a `BoxModuleConfig` wrapper, so `BaseIntegrationTest` reaches the class runnable via `getBxClass()`.

## Tests
Smoke on 1.17.0-snapshot: BedrockServiceTest 70, SuspendResume 22, coreMiddleware 17, FlightRecorder 14 all pass; ClaudeTest's 7 live-key cases fail as on `development`.

## Relation to other PRs
Independent. Part of the split of #271 / #273 requested by @lmajano; the sibling PRs (#170, #198, #248, #272) each stand alone against `development` and were tested on both 1.13 and 1.17.

### DIFF
--- a/box.json
+++ b/box.json
@@ -17,7 +17,7 @@
 		"boxlang"
 	],
 	"boxlang": {
-		"minimumVersion": "1.8.0",
+		"minimumVersion": "1.14.0",
 		"moduleName": "bxai"
 	},
 	"private": false,
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### ⚠️ Migration
+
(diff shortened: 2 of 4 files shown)


lmajano:
@copilot resolve the merge conflicts in this pull request

PR #275, opened 2026-09-07
FlightRecorder: storageCallback, observe mode, overridable _saveSnapshot (#170)

### LINKED ISSUE #170: FlightRecorderMiddleware: Support custom storage backends via callback / observe mode

dougcain:
## Summary
Closes #170. `FlightRecorderMiddleware` gets pluggable storage and an observe mode, plus the review fixes found while wiring it to streaming providers.

- **`storageCallback( tape, interaction )`** — called after every recorded interaction with a cheap snapshot of the tape (new `interactions` array, entries shared with the live tape; treat as read-only). Replaces "subclass and override" as the way to persist tapes elsewhere.
- **`mode: "observe"`** — records without ever replaying.
- **`_saveSnapshot()`** is public/overridable; `_isCapturing()` replaces five copies of the mode check. Invalid modes throw `FlightRecorder.InvalidMode`.
- **Streaming**: binary stream bodies are taped base64 with an `encoding` marker instead of being dropped; stream interactions are fingerprinted (model + message count) and never answered by the wrong taped call; a taped `dataPacket` is a `duplicate()` so later aliasing can't break the fingerprint.
- **Strict replay throws** `FlightRecorder.StreamReplayUnsupported` on an unmatched stream call instead of quietly going live (strict promises zero network). See the Migration note in the changelog.

## Tests
`FlightRecorderMiddlewareTest` 14 → 29. coreMiddleware 17, SuspendResume 22, OutputGuard 14 unchanged and green.

## Relation to other PRs
Independent of the other split PRs (only `FlightRecorderMiddleware.bx` + its test + changelog). Part of the split of #271 requested by @lmajano.


src/main/bx/models/middleware/core/FlightRecorderMiddleware.bx
```diff
@@ -89,8 +109,21 @@ class extends="BaseAiMiddleware" {
...
+	 * Optional callback invoked after each interaction is appended to the tape (record and
+	 * observe modes only): `storageCallback( tape, interaction )`. Must be a closure or
+	 * custom function; ignored otherwise. Exceptions thrown by the callback are caught and
+	 * logged — they never interrupt recording.
+	 */
+	property name="storageCallback" type="any" default="";
+
 	// ==================== Internal State ====================
 
+	/**
```
lmajano: Make this static

dougcain: Done in 8a78e70 (was 1ac08c7 before rebasing onto development) — `static { VALID_MODES = [...] }`, referenced as `static.VALID_MODES` (2 sites). FlightRecorderMiddlewareTest 30/30.

PR #277, opened 2026-09-07
Streaming tool loop + batch HITL resume: Cohere v1, Bedrock, Bedrock model families (#248)

### LINKED ISSUE #248: Bedrock and Cohere streaming have no tool-call support, so streaming HITL/batch approvals silently no-op

dougcain:
## Summary
Closes #248. Streaming tool calls and batch human-in-the-loop resume for **Cohere v1** and **Bedrock (InvokeModel API)**, more Bedrock model families, and the cross-provider tool-argument fixes that fell out of reviewing it.

### Streaming tool loop
- Cohere: `tool-calls-chunk` deltas accumulated per index, malformed streamed arguments rejected (`CohereStreamError`), the follow-up turn streamed from the returned tool calls, SSE transport inside `wrapLLMCall`.
- Bedrock (InvokeModel path): `content_block_start` / `input_json_delta` / `content_block_stop` accumulation, streaming tool loop, and OpenAI-shaped (`openai.gpt-oss-*`, DeepSeek, Qwen) plus Cohere Command R families wired with their own dialects.
- A deferred batch suspends **once** with a positional resume ledger; `resumeToolBatchStream()` resumes on the stream and emits the terminator once.
- `AiAgent.resumeStream()` refuses providers without streaming resume via `IAiService.supportsStreamingToolResume()` (structural default on `BaseService`, Gemini opts out).

### Tool-argument resolution (all providers)
- `BaseService::resolveToolArgs` replaces three per-provider copies. Comparison is canonical JSON: `duplicate().equals()` is **false** for any struct holding a JSON `null`, so such tool calls always looked "edited". A non-struct `ctx.toolArgs` falls back to the native args instead of invoking the tool with `{}`.
- `HumanInTheLoopMiddleware`'s edit decision writes **both** `ctx.toolArgs` and the native slot; a human edit was dropped whenever an earlier middleware had rewritten `toolArgs`.
- Claude / Cohere / OpenAI resume ledgers carry each entry's resolved `toolArgs`, so a suspended batch runs what the straight path would have. OpenAI previously never consulted `ctx.toolArgs` at all.

## Tests
BedrockServiceTest 93, CohereStreamingToolsTest 13, SuspendResumeIntegrationTest 36 (incl. Claude/Cohere/OpenAI ledger round trips + legacy ledger), HumanInTheLoopGateway 22, AiAgentResumeStreamGuard 6, ClaudeTest +3 — all green on both BoxLang 1.13 and 1.17. Live-key ClaudeTest/OpenAITest cases fail as on `development`.

## Relation to other PRs
Independent against `development`. The Bedrock **Converse** API PR (#272) is stacked on this one. Part of the split of #271 / #273 requested by @lmajano.


dougcain:
Split into three smaller PRs as requested, closing this one:
- #280 shared tool-arg resolution, ledger toolArgs, streaming-resume guard (independent against development, main +251/−64)
- #281 Cohere v1 streaming tool loop (stacked on #280, main +829/−123 in one file)
- #282 Bedrock InvokeModel streaming tool loop + model families (stacked on #281, main +1936/−357 in one file)

The tree at the top of #282 is byte-identical to this branch's head (`7297d9f`), so the review and test results here carry over. The Converse PRs (#283, #284) and #279 are stacked on #282.

PR #285, opened 2026-09-07
Code-review follow-ups for the #280–#284 / #279 stack

dougcain:
## Summary
Code-review follow-ups for the stack #280 → #281 → #282 → #283 → #284 → #279. **Stacked on #279**; review only the top commit (`6fb5b36`). Kept as one PR so the six reviewed commits stay byte-identical to what was reviewed and live-tested; every changelog bullet names the PR it corrects, so the fixes can be folded down into their originating PRs instead if you prefer — say so and I will.

Review: xhigh pass over the cumulative diff (11 finder angles, 8 verifier groups, gap sweep). 15 confirmed findings reported plus ~20 lower-severity confirmed items; all fixed here except the design notes listed under *Not changed*.

### Shared (#280 / #279)
- `resolveToolArgs` compared canonical JSON with case-insensitive `==`, so a case-only native-slot edit was dropped; now `compare()==0`, snapshot serialized once.
- OpenAI re-parsed `function.arguments` unguarded inside the `nativeArgs` closure (and ledger fallbacks), throwing out of `chatStream()` on `""`/truncated args; new guarded `toolCallArgs()` used everywhere.
- Claude's silent empty catch on the streamed tool-input parse now logs. Dead `capturedToolCall` / write-only `toolCallIndexMap` removed.
- MockService accepted `errorHolder` and ignored it; emissions now go through `emitStreamChunkGuarded` so mock streams announce `onAIError` and keep `onAITokenCount`.
- ClosureTool `_chatRequest` injection: keyed on the callable's real declared parameters (a hand-set `methodParameters` list can no longer suppress it), docblock reworded, **changelog bullet added for the opt-in behaviour change**, unit tests added.
- FlightRecorder `wrapLLMCall` is stream-aware: record passes through with a `{ stream: true }` marker, replay throws `FlightRecorderStreamingReplayUnsupported` instead of returning a mangled body / empty stream.
- `AiAgent` no longer writes `editedArgs = {}` on an edit with no data; every provider guards the slot with `isStruct`.
- Seven `chatStream()` overrides (Perplexity, HuggingFace, Grok, Groq, MiniMax, Mistral, OpenRouter) now pass `interactionCount` through (pre-existing).

### Cohere (#281)
- Follow-up rounds fire `beforeLLMCall` and `onAITokenCount` and apply `chat()`'s error-shape check (the removed recursion used to do all three).
- Stream transport: body built inside the `wrapLLMCall` inner function (in-place edits reach the wire), accumulators reset per attempt (RetryMiddleware safe), canned non-stream return rejected with a clear `CohereStreamError`.
- `finalizeToolBatchBuffers` no longer skips `afterLLMCall`; a `finish_reason: "stop"` terminator always follows the follow-up text; sparse tool schemas get defaults; catch-scope qualification.

### Bedrock (#282 – #284)
- Stream-callback parity: holder + `emitStreamChunkGuarded`, rethrow after `onAITokenCount` / `onAIChatResponse`; `BedrockStreamError` announce keyed on an "already announced" marker so malformed-frame / truncated-argument throws reach `onAIError`; `processBedrockEvent` rethrows `BedrockStreamError`; payload-truncated frames throw like header-truncated ones; `:error-message` header used for empty error frames.
- Converse request: `tool_choice: "none"` omits `toolConfig`; `response_format` dropped with a warning; `bedrockHeaders` shorthand filtered like top-level headers; one shared content-block key list (`guardContent` included); region-prefixed `openai.` ids get `max_completion_tokens`.
- Stream state: `message_delta` usage no longer zeroes prompt tokens; `finishReason` normalized for cohere/titan/llama/mistral; index-less OpenAI-shaped fragments keyed by id; streamed cancel appends the blocked `toolResult`; openai-family structured output on the streaming InvokeModel leg refused up front (was a silent no-result); fallback log keyed by `model|reason`; string-replay frames may carry empty `bytes`.
- `streamState` fully initialized, dead lazy-init and `isStruct` guards removed; stale class comments and orphaned docblocks fixed; catch-scope qualification (`local.truncated` was a real bug on ≤1.13).

### Not changed (design notes, left as-is)
- `resolveToolArgs` precedence (normalized slot wins once moved) is documented intent; HITL writes both slots.
- `countTokens( honourRuntimeDecision=false )` counts what a *fresh* logical call would send; mid-loop/resume counts may differ — documented in the docblock rather than re-plumbed.
- `PromptSecurity.getResponseText` fail-open on unknown shapes is pre-existing (this stack narrows it).
- `ClaudeService.resumeToolBatchStream` finishes via `chat()` (pre-existing).
- `countTokens` hand-rolled transport; deep `duplicate()` in the OpenAI-shaped transform (deliberate isolation).

## Tests
+35 offline regressions: BedrockServiceTest RF1–RF17 (179 total), CohereStreamingToolsTest +10 (25), ClaudeTest, StreamCallbackPropagationTest (9), ClosureToolTest (31), FlightRecorderMiddlewareTest (16), SuspendResumeIntegrationTest (43), AiAgentResumeStreamGuardTest (6). Full suite: 1777 tests, failure set identical to `development` (147 live-key / DB-backed tests). `testBedrockLive` 9/9 on eu-west-2.

Size: main +790/−316 in 16 files, tests +1736.


dougcain:
Folded into the originating PRs as requested (nothing had been reviewed yet): #280 `c79ae9b`, #281 `e820a9a`, #282 `85fbf1a`, #283 `03ca907`, #284 `761cb9d`, #279 `2b9efe1`. The tree at the top of the rebuilt stack is byte-identical to this PR's head (`6fb5b36`), so the full-suite and live results here carry over. Each PR body lists the fixes it absorbed. Closing.

